PR #2, opened 2026-07-22
feat: Compile static workflow graphs

lox:
The repository has the Phase 0 execution and trust foundations, but it cannot yet turn a supported GitHub Actions workflow into a native Buildkite job graph. This adds the Phase 1 static compiler so workflows can be checked, expanded, and inspected before any runtime or artifact transport is involved.

The compiler now builds deterministic compile-time contexts, expands matrices and repository-local reusable workflows, applies a fail-closed runner policy, produces versioned content-addressed job plans, and emits Buildkite pipeline YAML. `validate` also has stable text and JSON compatibility reports, while `compile --format ir-json` keeps the owned intermediate representation available for debugging.

For example, the shell smoke workflow becomes one producer and two matrix consumers with exact dependency fan-out. Generated jobs carry the expected plan digest and refuse dependency-bearing execution until Phase 2 provides producer-attributed result manifests, so this phase does not overstate the current runtime or transport boundary.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,0 +1,4 @@
+steps:
+  - label: ":go: Repository checks"
+    key: "checks"
+    command: "make check"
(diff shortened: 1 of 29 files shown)


sj26:

PR #4, opened 2026-07-22
test: Close Phase 0 with live transport evidence

lox:
Phase 0 still depended on dormant probes, so its transport and differential assumptions had never crossed real provider boundaries.

This makes the manual probes runnable on the existing unprivileged CI queue, adds a disposable plan-envelope signer for mechanics-only testing, and records exact hosted evidence for artifact transport, dependency extension, logical failure settling, signature rejection, Agent redaction, and interrupted-upload recovery. The default repository pipeline keeps both probes dormant unless `PHASE0_PROBE` explicitly selects one.

The closeout deliberately does not claim a production trust boundary. Buildkite signed pipelines move to optional Phase 9 defence in depth; KMS-backed plan signing, verification-only roots, checkout-free compiler isolation, and protected-capability policy remain required hardening gates.


sj26:

PR #5, opened 2026-07-23
feat: Add the sequential shell runtime

lox:
Phase 1 can compile supported GitHub Actions workflows into Buildkite jobs, but
those jobs previously had no production-shaped shell executor or trustworthy
way to carry `needs` results between generated steps. This adds the Phase 2
runtime and transport boundary so the shell smoke workflow can run end to end
on hosted Buildkite agents.

`buildkite-gha upload` now materializes the exact executable and
content-addressed plans, uploads a checkout-free pipeline, and binds every
generated job to its importer, build, step, queue, and plan digest. The unsigned
development path is deliberately narrow: it accepts only the fixed `hosted`
queue and rejects action steps, secrets, and every declared runtime capability.

`run-job` executes sequential Bash and sh steps in a fresh workspace with the
supported environment files, conditions, timeouts, cancellation,
`continue-on-error`, and masking behavior. Producers publish bounded,
producer-attributed result manifests; consumers verify those manifests before
using `needs.<job>.result` or outputs. For example, the checked-in shell fixture
now runs one producer and two matrix consumers, then releases a native
Buildkite continuation only after both consumers settle.

The plan records the live Buildkite proof and the intentional compatibility
limits, including whole-build recovery for ambiguous producer retries and the
current `::add-mask`-only workflow-command surface.


sj26:

PR #9, opened 2026-07-24
fix: align expression handling across boundaries

mcncl:
## Description

Fix three expression-boundary issues identified while testing Phases 1–3:

- Resolve reusable-workflow inputs in job and step conditions
- Support static bracket-form secret references such as `secrets['TOKEN']`
- Evaluate job-level `defaults.run` expressions at runtime

### Reusable-workflow conditions

- Substitute whole job and step conditions that directly reference static `inputs.*`
- Preserve input types when constructing conditions, including quoted string literals
- Support conditions with and without `${{ }}` delimiters
- Reject more complex residual `inputs.*` conditions at compile time instead of allowing them to fail at runtime

### Secret references

- Replace regex-based secret discovery with expression AST traversal
- Treat dot and static bracket forms equivalently
- Deterministically populate required secrets and the `secrets` capability
- Reject dynamic indexes and whole-context secret access
- Resolve bracket-form references at runtime and register their values for log masking

### Run defaults

- Apply explicit-step-over-job-default precedence first
- Evaluate the effective shell and working directory in the runtime expression context
- Cover `matrix`, `env`, and `vars` expressions for both default fields
- Verify explicit step settings avoid evaluating overridden defaults

## Verification

- `go test ./...`
- `go test -race ./internal/expression ./internal/compiler ./internal/runtime`
- `go vet ./...`
- `golangci-lint run`
- `git diff --check`

### DIFF
--- a/internal/compiler/bundle_test.go
+++ b/internal/compiler/bundle_test.go
@@ -105,7 +105,7 @@ func TestCompileBundleDoesNotExposeEventValues(t *testing.T) {
 }
 
 func TestCompileBundleDeclaresSecretCapabilityAndNames(t *testing.T) {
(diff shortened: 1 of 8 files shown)


sj26:

PR #10, opened 2026-07-24
fix(runtime): make explicit cancellation deterministic

mcncl:
## Description

- observe direct-shell exit independently from stdout and stderr draining
- stop the shell while interrupting its same-process-group children, then resume it so its pending signal trap can run
- wait for the direct shell to commit file-command effects before terminating remaining descendants
- add Linux and Darwin process-group enumeration
- strengthen the explicit-cancellation regression test

## Why

Previously, cancellation sent `SIGINT` to the entire process group at once. Under scheduling contention, the shell's foreground child could exit while Bash was processing the same signal, causing Bash to terminate without consistently running its `SIGINT` trap.

This was observable as intermittent missing trap effects, including GitHub file-command updates that should have been committed during graceful cancellation.

## Testing

- [x] `mise run check`
- [x] `go test ./...`
- [x] `go test -race ./internal/runtime`
- [x] `go vet ./...`
- [x] `golangci-lint run`
- [x] Run the focused cancellation contract 1,000 times
- [x] Run the focused cancellation contract 1,000 times with the race detector
- [x] Cross-build the runtime package for Linux and Windows

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/rhysd/actionlint v1.7.12
 	github.com/santhosh-tekuri/jsonschema/v6 v6.0.2
 	go.yaml.in/yaml/v4 v4.0.0-rc.3
+	golang.org/x/sys v0.42.0
 )
 
 require (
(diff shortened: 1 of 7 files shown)


sj26:

PR #22, opened 2026-07-27
Add durable Docker orb setup

lox:
## Summary

- install `docker.io` without starting its package-managed Docker or containerd units
- run the rootful daemon exclusively through Amp's service supervisor, exposing the conventional `/var/run/docker.sock` to the orb user's group
- use Docker's persistent `/var/lib/docker` data directory, default context, native `overlay2` storage, and systemd cgroups
- install checksum-pinned Docker Buildx v0.35.0 and keep its default builder on the local Docker driver
- install the Buildkite CLI v3.54.1 and GitHub CLI v2.96.0 globally through mise, with login-shell shim activation

## Validation

- repeated `.agents/setup` and `.agents/resume` runs
- supervised daemon stop/resume with image persistence
- `docker info`: Docker 20.10.24, `overlay2`, systemd cgroups, `/var/lib/docker`
- `docker run --rm alpine:3.20 ...`
- `docker buildx inspect default` reports driver `docker`
- host-gateway HTTP access from the default bridge and a user-created network
- `BUILDKITE_GHA_LIVE_REQUIRED=1 go test ./internal/runtime -run '^TestLiveJobContainerJavaScriptCompositeAndPost$' -count=1 -v`
- fresh login shell resolves mise, `bk version 3.54.1`, and `gh version 2.96.0`
- Bash syntax, ShellCheck, and `git diff --check`

Probe containers, networks, listeners, and services were removed after validation. The package-managed Docker, socket, and containerd units remain disabled and inactive; only the Amp-supervised Docker service remains running.

The Docker socket grants root-equivalent access inside the dedicated per-thread orb. Test containers are not given that socket.

### DIFF
--- a/.agents/resume
+++ b/.agents/resume
@@ -1,4 +1,96 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-echo "No persistent services require restart."
+uid="$(id -u)"
+runtime_dir="/tmp/user-$uid"
+docker_data_dir="$HOME/.local/share/docker-rootless"
+docker_context="buildkite-gha-rootless"
+docker_endpoint="unix://$runtime_dir/docker.sock"
+default_socket="/var/run/docker.sock"
+rootless_socket="$runtime_dir/docker.sock"
+
+for command in amp docker rootlesskit slirp4netns fuse-overlayfs; do
+  if ! command -v "$command" >/dev/null; then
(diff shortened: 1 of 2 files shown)


lox:
Superseded by #23, which uses the corrected `lox/orb-docker-setup` branch and a single clean rootful-Docker setup commit.

PR #24, opened 2026-07-27
ci: run Buildkite pipeline on hosted agents

lox:
## Summary

- make `hosted` the root queue for every checked-in Buildkite pipeline document
- remove the control-plane dependency on the unreliable `elastic-runners` queue while retaining explicit generated runtime queue bindings
- update the dormant transport probe instructions and add an invariant test covering all checked-in pipeline YAML and rejecting non-hosted step overrides

This avoids the elastic runner failure mode seen in [Buildkite build 190](https://buildkite.com/buildkite/buildkite-gha/builds/190/list?sid=019fa5ca-243e-44dd-abd4-eeb517a7127d&tab=output).

## Testing

- `mise run --jobs 1 check`
- `go test ./internal/buildkite` after addressing Oracle review

## Oracle review

Oracle recommended approval and found one test-hardening issue: a future step-level queue override could bypass the root queue invariant. Commit `34561ed` now rejects any checked-in non-hosted step override.

## Rollout checks

- ensure the Buildkite UI-defined bootstrap/pipeline-upload step also targets `hosted`; repository YAML cannot select the agent that uploads the root document
- confirm `GHA_GITHUB_RELEASE_TOKEN` remains available to the tagged release job on the hosted queue before the next release
- disable the elastic queue only after a complete hosted build passes

### DIFF
--- a/.buildkite/hosted-smoke-aggregator.yml
+++ b/.buildkite/hosted-smoke-aggregator.yml
@@ -1,3 +1,6 @@
+agents:
+  queue: "hosted"
+
 steps:
   - label: ":white_check_mark: Hosted smoke targets settled"
     key: "hosted-smoke-terminal"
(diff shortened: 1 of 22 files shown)


sj26:

PR #25, opened 2026-07-28
test: avoid race detector sleep in Docker fakes

lox:
## Summary

- disable the race detector's exit sleep only for fake-Docker helper processes
- keep race instrumentation and the production subprocess environment allowlist unchanged
- reduce the full race suite from roughly nine minutes to under thirty seconds locally

## Why

The container tests repeatedly execute their own race-instrumented test binary to fake Docker commands. Go's race runtime sleeps for one second at process exit by default, so hundreds of helper invocations dominated the Repository checks step.

## Verification

- `go test -race ./... -count=1` (26.41s)
- `go test ./internal/runtime -count=1` (14.36s)
- `go test -race ./internal/runtime -count=1` (28.26s, previously ~529s)
- `git diff --check`
- `gofmt -d internal/runtime/containers_test.go`

Oracle reviewed the change and recommended approval with no blocking findings.

### DIFF
--- a/internal/runtime/containers_test.go
+++ b/internal/runtime/containers_test.go
@@ -46,7 +46,7 @@ func newJobDocker(t *testing.T, scenario string) fakeJobDocker {
 		t.Fatal(err)
 	}
 	wrapper := filepath.Join(root, "docker")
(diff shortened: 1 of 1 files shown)


sj26:

PR #28, opened 2026-07-30
fix: serialize repository checks

mcncl:
## Summary

- run the aggregate repository checks serially so standard and race-enabled Docker suites cannot observe each other’s resources
- require live Docker and managed Node prerequisites in the hosted repository check
- assert and document the hosted-versus-local prerequisite behavior

## Why

`mise run check` previously scheduled `test` and `test:race` concurrently. Their daemon-wide Docker leak assertions could see resources owned by the other test process and report false leaks. Hosted checks could also silently skip Phase 5 coverage when a live prerequisite was unavailable.

## Verification

- `mise exec -- go test ./internal/buildkite`
- `mise exec -- go clean -testcache && BUILDKITE_GHA_LIVE_REQUIRED=1 mise run check`
- confirmed no `buildkite-gha-*` Docker containers, networks, or images remained

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -76,6 +76,8 @@ steps:
 
   - label: ":go: Repository checks"
     key: "checks"
+    env:
+      BUILDKITE_GHA_LIVE_REQUIRED: "1"
     plugins:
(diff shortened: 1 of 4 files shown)


sj26:

PR #30, opened 2026-07-30
Bound aggregate job summaries

mcncl:
## Summary

- route pre, main, background, composite, nested, and post-action summaries through one deterministic accumulator
- cap the final job summary at 1 MiB with UTF-8-safe truncation and a clear notice
- preserve secret redaction safety when masking expands or intersects the truncation boundary
- treat an oversized per-step summary as a non-fatal upload failure while preserving the step's other effects

## Why

Job summaries were individually bounded but could grow without limit while being combined across a job. Establishing this runtime contract separately keeps the follow-up Buildkite annotation publication focused on presentation and transport.

## Testing

- `mise run check`


sj26:

PR #38, opened 2026-07-31
Fix Phase 6 artifact roundtrip proof workflow root

mcncl:
## Summary

The first hosted artifact-roundtrip proof failed in its importer with:

```
buildkite-gha: upload: build plan for job "artifact-producer": workflow path must identify a repository root
```

Nonce injection wrote the generated workflow under the external distribution temp directory. The compiler intentionally recognizes repository workflows only under a `.github/workflows` directory, so action-lock construction had no repository root.

This change:

- creates the generated workflow under the fixture's checked-in `.github/workflows` directory
- retains the external private directory for the built distribution
- guarantees cleanup of both temporary paths through the importer exit trap
- pins this path and cleanup behavior in the pipeline contract test

## Verification

- reproduced the importer compile locally with an injected build nonce and all three jobs
- confirmed the generated workflow is removed after the command exits
- `mise exec -- go test ./internal/buildkite`
- `mise run lint:shell`
- `mise run check`


sj26:

PR #41, opened 2026-08-03
Add Phase 6 cache roundtrip proof

mcncl:
## Summary

- expand the exact `actions/cache` v6.1.0 fixture into a build-unique producer/consumer miss, post-save, and exact-hit roundtrip
- add an exact-commit targeted importer, failure-tolerant settlement step, and independent build/job/log verifier
- keep the proof out of the aggregate hosted suite and the fixture at `compile-pass` while job-bound GHAC token minting remains disabled
- document the eventual targeted dispatch and verification procedure

## Verification

- `mise run check`
- `mise run smoke:profile`
- `git diff --check`

No hosted runtime evidence is claimed or attempted in this PR because token minting is not enabled yet.

### DIFF
--- a/.buildkite/phase-6-cache-roundtrip-continuation.yml
+++ b/.buildkite/phase-6-cache-roundtrip-continuation.yml
@@ -1,0 +1,12 @@
+agents:
+  queue: "hosted"
+
+steps:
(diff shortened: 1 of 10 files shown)


sj26:

PR #46, opened 2026-08-04
feat: expose reusable workflow outputs

mcncl:
## Summary

- parse declared `on.workflow_call.outputs` and expose direct `jobs.<job>.outputs.<name>` mappings through `needs.<call>.outputs`
- project only exact declared producer outputs into schema-v5 plans, preserving undeclared-output isolation and status-only inherited caller prerequisites
- compose declared outputs through nested local reusable calls and every exact matrix producer
- fail closed with source-located diagnostics for invalid, missing, ambiguous, or oversized mappings

## Contract

Supported declarations are one complete static reference:

```yaml
on:
  workflow_call:
    outputs:
      published:
        value: ${{ jobs.build.outputs.result }}
```

Literal and compound mappings remain deferred. Matrix-selected values verify every exact producer and fail when values conflict because the artifact contract has no authoritative completion order. A call remains bounded to 64 concrete output projections.

## Validation

- `mise run check`


sj26:

PR #47, opened 2026-08-04
test: prove reusable workflow outputs end to end

mcncl:
## Summary

- publish an exact declared output from the reusable quality workflow
- consume and assert that output in the dependent advanced migration POC job
- mark the changed advanced fixture compile-pass until a fresh exact-commit hosted run succeeds

This extends the existing full migration demo rather than adding an isolated proof or a hard-coded job count. The flow now fails if reusable workflow output projection or hydration regresses.

## Validation

- `mise run check`


sj26:

PR #50, opened 2026-08-04
docs: define Phase 6 capability grant boundary

mcncl:
## Summary

- propose the cross-repository trust contract for Phase 6 protected capabilities
- keep the Buildkite-owned control plane responsible for Job OIDC verification, independent GitHub provenance, policy, signing, and audit
- define a signed empty-capability grant as the first proof, returning no credential
- bind the request endpoint and OIDC audience to one installer-owned HTTPS origin
- use standard compact ES256 JWTs, explicitly separate from Phase 0 signing and cache credentials
- keep public tokenless workflows independent of control-plane availability

The ADR deliberately stops before private checkout, secrets, GitHub tokens, environments, or compatibility OIDC. Those remain separate reviewed slices after the no-op boundary is proven.

## Platform confirmations needed

- which Buildkite service/repository owns the endpoint
- whether authoritative GitHub webhook records can be joined to builds with enough fidelity to reproduce the canonical event snapshot
- how the corresponding audit record will be queried for independent hosted proof
- which immutable cluster/queue IDs are observable inside the runtime versus enforceable only during control-plane OIDC verification

## Validation

- `git diff --cached --check`
- current Buildkite Agent OIDC command and optional claims checked against the official docs and `buildkite/agent` source
- independent security-boundary review applied before opening

### DIFF
--- a/docs/architecture/0003-protected-capability-control-plane.md
+++ b/docs/architecture/0003-protected-capability-control-plane.md
@@ -1,0 +1,298 @@
+# ADR 0003: Authenticate protected capability grants through a Buildkite control plane
+
+- Status: Proposed
+- Date: 2026-08-04
+- Decision owners: `buildkite-gha` and Buildkite platform maintainers
+
+## Context
+
+The service-free compatibility path can execute public, anonymous workflow code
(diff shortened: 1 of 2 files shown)


sj26:

PR #51, opened 2026-08-04
docs: reconcile beta demo readiness

mcncl:
## Summary

- distinguish the hosted-proven runtime from the still-unproven public plugin installation path
- define a reproducible service-free demo baseline and an opt-in cache extension
- sequence the CLI v0.2.0, companion-plugin proof/tag, and final published quick-start evidence
- call out the remaining public-beta scope decisions without blocking tokenless demo hardening

## Verification

- `git diff --check`

## Notes

This is a plan-only reconciliation. It does not broaden the protected-capability boundary or claim that the current development cache endpoint is a public service contract.


sj26:

PR #53, opened 2026-08-04
test: make runtime suite portable on macOS

mcncl:
## Summary

- canonicalize test-owned temporary paths before asserting runtime and mise symlink behavior on Darwin
- make fake Docker and container execution portable across GNU/Linux and macOS utilities and host layouts
- probe live Docker with the same isolated configuration used by runtime execution, skipping locally when that secure path is unavailable

Runtime security behavior is unchanged: Docker credentials and connection overrides remain isolated, and managed mise paths retain strict symlink rejection.

## Verification

- `mise run check` (Linux amd64)
- `GOOS=darwin GOARCH=arm64 go test -c ./internal/runtime`
- `mise exec -- go test -count=1 ./internal/runtime` (macOS arm64, 50.753s)


sj26:

PR #59, opened 2026-08-04
Add paired Actions and Buildkite examples

lox:
## Summary

- promote the basic, artifact, and advanced migration POCs into manually dispatchable root workflows shared by both providers
- add a dedicated Buildkite picker pipeline that imports exactly one allowlisted workflow through the released github-actions plugin
- add an exact-commit launcher that starts matching GitHub Actions and Buildkite runs and prints both URLs
- keep the existing smoke inventory, migration importers, plugin demo, tests, and documentation aligned with the canonical workflow paths

## Validation

- `mise run check`

## Rollout

The manual [buildkite-gha Examples](https://buildkite.com/buildkite/buildkite-gha-examples) pipeline is already provisioned and points at `.buildkite/examples.yml`. GitHub only accepts `workflow_dispatch` for a new workflow after that workflow exists on the default branch, so the first paired run should be launched after this PR merges.

### DIFF
--- a/.buildkite/examples.yml
+++ b/.buildkite/examples.yml
@@ -1,0 +1,73 @@
+agents:
+  queue: "hosted"
+
+steps:
+  - block: ":github: Choose an example workflow"
+    key: "choose-example"
+    if: build.env("EXAMPLE") == null
+    prompt: "Choose the GitHub Actions workflow to run as native Buildkite jobs."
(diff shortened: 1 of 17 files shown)


sj26:
Released in [v0.2.1](https://github.com/buildkite/buildkite-gha/releases/tag/v0.2.1)

PR #60, opened 2026-08-04
test: canonicalize Darwin temp paths

mcncl:
## Why

`mise run release` correctly runs the full check gate before tagging, but two tests compare logical macOS `/var/...` temp paths with production paths canonicalized through `/private/var/...`. This makes a clean patch release fail locally on Darwin even though the production behavior is correct.

## What

- canonicalize the action metadata test root before asserting the exact source-located diagnostic
- canonicalize the mise cache test roots so the success case is a real directory and the symlink case tests the explicit fixture symlink rather than macOS's system `/var` alias
- leave production path and symlink checks unchanged

## Validation

- `mise exec -- go test -count=1 ./internal/action/metadata ./internal/cli`
- `mise run check`
- `git diff --check`


sj26:
Released in [v0.2.1](https://github.com/buildkite/buildkite-gha/releases/tag/v0.2.1)

PR #61, opened 2026-08-05
Add generated example pipeline uploader

lox:
## Summary

- add an executable Buildkite bootstrap that uploads the selected example importer directly when `EXAMPLE` is set
- generate the interactive picker only for manual builds without a selection
- validate the fixed example allowlist and exact Buildkite commit before uploading
- retain `.buildkite/examples.yml` during the pipeline-settings rollout

## Testing

- `mise run check`

### DIFF
--- a/.buildkite/upload-examples.sh
+++ b/.buildkite/upload-examples.sh
@@ -1,0 +1,107 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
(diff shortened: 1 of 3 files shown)


sj26:
Released in [v0.2.2](https://github.com/buildkite/buildkite-gha/releases/tag/v0.2.2)

PR #65, opened 2026-08-05
Resolve matrix expressions in job labels

lox:
## Summary

- evaluate explicit workflow job names against each concrete matrix row before emitting Buildkite labels
- avoid appending redundant matrix metadata when the explicit name already depends on matrix values
- preserve the existing matrix suffix for static or non-matrix-dependent names

This was found by running the merged workflow UX from source in the examples pipeline: the advanced jobs appeared as `Verify payload (${{ matrix.variant }}) (variant=primary)`. They now appear as `Verify payload (primary)` and `Verify payload (secondary)`.

## Validation

- `mise run check`
- source-built advanced example: https://buildkite.com/buildkite/buildkite-gha-examples/builds/9

### DIFF
--- a/internal/compiler/compiler.go
+++ b/internal/compiler/compiler.go
@@ -634,7 +634,7 @@ func expand(path string, source []byte, parsed *workflow.Workflow, context expre
 			instance := JobInstance{
 				Key:                     key,
 				LogicalJobID:            job.ID,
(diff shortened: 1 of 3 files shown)


sj26:
Released in [v0.2.3](https://github.com/buildkite/buildkite-gha/releases/tag/v0.2.3)

PR #72, opened 2026-08-06
Harden credential service boundaries

mcncl:
## Summary

- restrict `buildkite-agent redactor add` to the Job API socket and token instead of inheriting the complete job environment
- require HTTPS for GitHub-token, cache-token, and cache Results service endpoints, while retaining literal loopback HTTP for local testing
- add regression coverage for allowed IPv4/IPv6 loopback endpoints and rejected hostname, private-IP, and public-IP plaintext endpoints

## Security impact

This reduces the credentials available to the redactor subprocess and prevents bearer credentials from being sent to non-loopback services over plaintext HTTP. Existing URL protections for userinfo, queries, fragments, and cache Results paths remain enforced.

## Testing

- `mise run check`

### DIFF
--- a/internal/runtime/cache_service.go
+++ b/internal/runtime/cache_service.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/url"
 	"regexp"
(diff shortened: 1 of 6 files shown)


sj26:
Released in [v0.4.1](https://github.com/buildkite/buildkite-gha/releases/tag/v0.4.1)

PR #73, opened 2026-08-06
Fail unsupported conditions during validation

mcncl:
## Summary

- statically validate workflow job and step `if` conditions against the expression forms and runtime contexts implemented by buildkite-gha
- reject unsupported functions, operators, reference shapes, dynamic indices, and unavailable contexts before generated jobs are uploaded
- report the workflow path, exact condition location, affected job, and affected step where applicable
- update the public compatibility documentation to describe the preflighted condition subset

## Supported behavior preserved

Runtime-dependent conditions using `needs`, `steps`, `matrix`, `vars`, step `env`, service ports, retained GitHub identity fields, and the zero-argument status functions continue to compile unchanged. Local reusable-workflow inputs are still statically substituted before condition validation.

The validation is shared by `validate`, profile validation, `compile`, and `upload`. The upload test verifies an incompatible condition fails before any Buildkite Agent command is invoked.

## Scope

This does not add expression functions, transport the event payload into job plans, or change workflow/job concurrency handling.

## Validation

- `mise run check`
- independent review for compiler/runtime phase mismatches and false positives

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,8 @@ The plugin path currently supports:
 - static matrices, ordinary `needs` and outputs, and local reusable workflows
   with statically resolved inputs, caller-visible results, and directly mapped
   declared outputs;
-- the currently implemented job and step condition subset;
+- the documented job and step condition subset, with unsupported functions and
+  unavailable runtime contexts rejected before pipeline upload;
 - background, wait, cancellation, and parallel step controls;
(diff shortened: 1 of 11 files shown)


sj26:
Released in [v0.5.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.5.0)

PR #85, opened 2026-08-08
Speed up runtime tests with safe parallelism

lox:
## Problem

The test suite is dominated by `internal/runtime`, where process-heavy runtime and fake-Docker tests all ran serially. A warm uncached run took about 16 seconds normally and 32 seconds under the race detector.

## Change

- run 32 independent runtime, container, cancellation, and process-helper tests in parallel
- keep tests that mutate process-wide environment variables serial
- scope fake-Docker root and scenario state to each subprocess instead of the parent test process, removing the shared state that prevented safe parallel execution

This changes only the test harness. Production behavior and coverage are unchanged.

## Results

Measured in the same 16-core Amp orb with test-result caching disabled:

| Suite | Before | After |
| --- | ---: | ---: |
| `go test ./...` | 16.33s | 8.05s |
| `go test -race ./...` | 31.95s | 18.46s |

That makes the normal suite 51% faster and the race suite 42% faster.

## Validation

- `go test -count=10 ./internal/runtime`
- `go test -race -shuffle=on -count=3 ./internal/runtime`
- `mise run check` after rebasing onto current `main` (39.60s, 0 lint issues)

### DIFF
--- a/internal/runtime/containers_test.go
+++ b/internal/runtime/containers_test.go
@@ -48,13 +48,13 @@ func newJobDocker(t *testing.T, scenario string) fakeJobDocker {
 		t.Fatal(err)
 	}
 	wrapper := filepath.Join(root, "docker")
-	// Avoid the race runtime's one-second exit sleep on every fake Docker call.
+	// Scope fake state to this subprocess rather than the parent test process so
+	// independent container tests can run in parallel. Avoid the race runtime's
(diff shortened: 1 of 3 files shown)


sj26:
Released in [v0.6.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.6.0)

PR #91, opened 2026-08-08
Reuse scoped job tokens for hosted public action sources

lox:
## Summary

Revises the hosted, zero-configuration authentication design for public GitHub Action sources to reuse Buildkite's existing job-bound scoped-token endpoint rather than add a dedicated public-source endpoint and GitHub App.

The ADR now distinguishes the observable GitHub Actions architecture: the runner normally delegates resolution to a job-bound Actions/Launch service and prefers resolver-supplied per-action download authentication, but explicitly falls back to the job's workflow-repository `GITHUB_TOKEN` when no download token is supplied. The current `buildkite-gha` design matches that fallback shape; it does not claim parity with GitHub's full server-side action resolver.

This PR remains design-only. It does not enable hosted authenticated resolution and contains no runtime or client implementation.

## Proposed contract

- Reuse `POST /v3/jobs/<current-job-id>/github_scoped_access_token`, authenticated with the current job's Agent token.
- The importer constructs a fixed request for the exact authoritative pipeline/event repository and exactly `contents: read`; workflow content controls neither field.
- Backend authorization must bind the job and pipeline repository, limit selected installation repositories to that exact repository, grant no write or other configurable permissions, expire within one hour, and use the customer's installation/rate bucket.
- `BUILDKITE_GHA_GITHUB_TOKEN` remains an explicit non-hosted local/corpus fallback only. Hosted `upload` ignores it, prefers the Agent endpoint, and falls back anonymously on bounded availability failures.

## Security properties

- Registers the token with the Agent redactor before any use.
- Keeps it in compiler memory only; it never enters plans, generated jobs, workflow contexts, action inputs, subprocess environments, arguments, URLs, caches, logs, metrics, or traces.
- Requires each distinct action repository's authenticated metadata to report `private: false` before refs, annotated tags, commits, or tarballs are requested.
- Sends authorization only to HTTPS `api.github.com` and strips authorization and cookies before expected redirects to `codeload.github.com`.
- Treats access to an exact private pipeline repository as expected endpoint scope, not public-action authority. Public-only behavior is enforced by compiler containment and the metadata gate.
- Never consults ambient `GH_TOKEN`/`GITHUB_TOKEN`, workflow-provided tokens, PATs, or repository credential helpers for hosted source resolution.

## External backend work

Backend owners must confirm or add importer caller support and contract tests proving:

- same-job Agent authorization and exact authoritative pipeline-repository binding;
- one selected installation repository rather than all installation repositories;
- requested `contents: read`, GitHub's automatic `metadata: read`, and no write or other configurable permission;
- lifetime no longer than one hour;
- customer installation and authenticated rate-limit bucket behavior; and
- sanitized failures with no Agent token, installation token, or upstream body in logs, traces, metrics, or errors.

Hosted live conformance must cover public metadata, lightweight and annotated refs/tags, commits, and tarballs outside installation selection; private and write negatives; redirect credential stripping; and end-to-end non-leakage.

## Validation

- `mise run format`
- `git diff --check origin/main...HEAD`
- Confirmed `origin/main...HEAD` contains only `docs/architecture/0004-public-action-source-authentication.md`.

## Non-goals

- Enabling hosted authenticated Action resolution in this PR.
- Adding runtime client or compiler wiring.
- Adding `github_public_source_token` or a dedicated zero-permission App now; the ADR records explicit reconsideration triggers.
- Claiming parity with GitHub's job-bound action resolver or per-action download tokens.
- Adding private Action source support. A future job-bound resolver with exact per-action credentials is the preferred direction for that scope.
- Adding PAT flags or consulting ambient `GH_TOKEN`/`GITHUB_TOKEN` in hosted builds.
- Refactoring the general Action resolver before backend coordination.

### DIFF
--- a/docs/architecture/0004-public-action-source-authentication.md
+++ b/docs/architecture/0004-public-action-source-authentication.md
@@ -1,4 +1,4 @@
-# ADR 0004: Authenticate hosted public Action source resolution with a dedicated GitHub App
+# ADR 0004: Reuse the scoped job token for hosted public Action source
 
 - Status: Proposed
 - Date: 2026-08-08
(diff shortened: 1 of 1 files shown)


lox:
Closing at Lachlan’s direction. This ADR draft is superseded by a minimal implementation that will reuse the existing job-scoped token endpoint directly, without a new environment variable or separate architecture PR.

PR #92, opened 2026-08-08
Authenticate public action-source resolution with BUILDKITE_GHA_GITHUB_TOKEN

lox:
## Summary

- add optional importer-only authentication for public remote action-source metadata, ref, commit, and tarball requests using `BUILDKITE_GHA_GITHUB_TOKEN`
- preserve anonymous resolution when the variable is missing or empty; `GH_TOKEN` and `GITHUB_TOKEN` are not fallbacks
- document safe importer-process configuration for corpus and production-policy testing

## Security boundary

- repository visibility is checked independently before authenticated resolution and materialization, including cache hits, so a broad token does not enable private action sources
- private and missing repositories retain the same not-found-or-not-public error semantics
- authorization is sent only to the trusted GitHub API origin and is stripped, along with cookies, before redirects to codeload or another permitted archive final host
- the credential is held only in source-client configuration and is never serialized into plans, generated pipeline YAML, workflow contexts, action inputs, action subprocess environments, caches/manifests, or logs
- operators must configure this credential only in the importer process environment, not at Buildkite pipeline or build scope, because buildkite-gha cannot undo Buildkite-level environment propagation to generated jobs

## Validation

- `mise run check`
- `git diff --check origin/main...HEAD`

## Non-goals

- enabling private action repositories
- providing workflow `GITHUB_TOKEN` or any workflow-visible credential
- invoking `gh auth token` or accepting token literals on the command line
- changing runtime action credentials or private checkout behavior


lox:
Closing at Lachlan’s direction. The replacement will retain only the necessary authenticated resolver hardening and wire it to the existing job-scoped token endpoint; BUILDKITE_GHA_GITHUB_TOKEN will not be introduced.

PR #93, opened 2026-08-09
Run node20 actions with Node 24

lox:
## Why

GitHub-hosted runners now execute actions declaring `runs.using: node20` with their bundled Node 24 runtime by default. buildkite-gha still selected its exact Node 20 runtime, creating a compatibility mismatch.

## What

Map `node20` declarations to buildkite-gha's existing digest-verified Node 24 runtime at the shared metadata classification boundary, including composite, remote lifecycle, and lazy job-container paths. Explicit `node24` declarations remain unchanged, and unresolved workspace actions no longer cause Node 20 to be mounted speculatively.

The temporary `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION` escape hatch is intentionally unsupported: buildkite-gha has no server-controlled removal phase, so exposing it would make the insecure fallback indefinite rather than matching GitHub's bounded transition.


sj26:

PR #95, opened 2026-08-10
Validate action input defaults during compilation

wolfeidau:
## Why

Resolved actions can currently pass compilation even when an effective input default uses expression syntax that the runtime cannot evaluate. This defers a compatibility failure until `run-job`, after the pipeline has been uploaded, and surfaces a confusing step-level error.

## What

Validate effective action input defaults during compilation and reject unsupported runtime interpolation before plans or pipeline YAML are emitted. Explicitly supplied inputs continue to bypass unused defaults.

## How

- Add a non-evaluating runtime-template validator backed by the same reference classifier used by runtime evaluation.
- Inspect resolved local, remote, and nested composite action invocations while building action requirements.
- Fail upload before any Buildkite calls when an effective default is unsupported.
- Add expression, compiler, remote-action, override, and CLI regression coverage.

### DIFF
--- a/internal/cli/cli_test.go
+++ b/internal/cli/cli_test.go
@@ -410,6 +410,51 @@ jobs:
 	})
 }
 
+func TestUploadRejectsUnsupportedEffectiveActionInputDefaultBeforeBuildkiteCalls(t *testing.T) {
+	root := t.TempDir()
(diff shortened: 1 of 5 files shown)


lox (approved):

sj26:

PR #97, opened 2026-08-10
Isolate upload test from tag builds

lox:
A tag release build inherits `BUILDKITE_TAG`, which caused the upload webhook test to assert a tag ref instead of the branch ref configured by the test. Clear `BUILDKITE_TAG` in that test, matching the neighboring derived-event test, so repository checks are deterministic in both branch and tag builds.

This unblocks publishing a replacement `v0.7.1` release while leaving the failed `v0.7.0` tag immutable.


sj26:

PR #98, opened 2026-08-10
Add pinned open-source workflow compatibility corpus

lox:
## Why

The repository-owned smoke fixtures prove specific contracts, but they do not show where unchanged real-world workflows compile, pass admission, or fail closed. We need broader compatibility evidence without executing unreviewed third-party code or adding network access to the normal check gate.

## What

- add ten workflows identified by exact public repository commit, path, and raw-file SHA-256, with seven exact-SHA successful GitHub run references
- add an opt-in scanner that downloads only each workflow YAML, verifies its digest, constructs a minimal event, and reports each case as passing or blocked
- run the full `hosted-tokenless` profile scan in one Buildkite step, publish a per-case annotation, and soft-fail the step while compatibility gaps remain
- keep `mise run check` network-free; neither scanner mode clones repositories or executes action or repository code
- deliberately defer repository-local actions/reusable workflows and all Hosted execution until a selected case requires a separately reviewed extension

## Current result

- compile: **7 passing, 3 blocked**
- hosted-tokenless admission: **1 passing, 9 blocked**
- only `bat-changelog` currently reaches admission
- five workflows expose the same unsupported compound `setup-go`/`setup-node` token default
- `jq-valgrind` exposes an unaudited `upload-artifact@v7` adapter commit
- the remaining three fail closed at compile time for dynamic concurrency, reusable-workflow secrets, or a macOS matrix entry

The Buildkite canary exits non-zero whenever a case is blocked, so the job is visibly soft-failed rather than presenting known incompatibilities as successful expected classifications.


sj26:

PR #100, opened 2026-08-10
Improve OSS action compatibility

lox:
## Why

The OSS compatibility canary exposed two bounded action-adapter gaps that can be fixed without changing workflow meaning:

- `actions/upload-artifact@v7` resolves to `043fb46d1a93c77aae656e7c1c64a875d1fc6a0a`, while the native adapter admitted only its audited v4 commit
- `actions/checkout` rejected `fetch-depth: 0`, used by the pinned fzf workflow

## What

- admit the audited upload-artifact v7 commit alongside the existing v4 commit
- retain the existing bounded ZIP-only artifact contract and continue rejecting `archive: false` explicitly
- support checkout depth 0 by fetching all branch and tag refspecs plus a private ref pinned to the exact event SHA
- retain depth 1 as the default and continue rejecting arbitrary positive depths
- test both audited artifact pins, unknown-pin rejection, v7 ZIP/raw boundaries, checkout input classification, and exact full-history Git arguments
- document both boundaries

This removes both failures without pretending that a SHA-only fetch is GitHub-compatible full history and without silently treating v7 raw uploads as ZIP uploads.

## Corpus impact

With the separately-owned compound action-default expression fix simulated by explicit empty token inputs, the unchanged fzf workflow reaches `admitted` with its real `fetch-depth: 0` setting.

The `jq-valgrind` case advances beyond the v7 commit check but remains correctly blocked on independent choices: checkout submodules, artifact glob paths, and per-artifact retention.

## Validation

- `mise exec -- go test ./internal/action/integration ./internal/compiler ./internal/runtime -run 'Test(UploadArtifact|Checkout)' -count=1`
- `mise run check`
- real Git fetch against `junegunn/fzf@715d26f`: 11 remote branches, 177 tags, and exact event ref verified
- exact fzf profile probe: `admitted` after bypassing only the separately-owned compound token-default parser issue


sj26:

PR #101, opened 2026-08-10
Support toolchain images for generated jobs

lox:
## Why

The new opt-in Hosted toolchain images provide an image-baked `/opt/hostedtoolcache`, but generated `run-job` steps cannot currently select those images or expose their trusted tool inventory to setup actions. Falling back to a persistent writable tool cache would weaken the executable-cache boundary because setup actions trust entries and completion markers without verifying their contents.

## What

- let importers select one runtime image with `BUILDKITE_GHA_RUNTIME_IMAGE`
- require an immutable registry digest and emit it as the `image` for every generated workflow job
- point `RUNNER_TOOL_CACHE` at `/opt/hostedtoolcache` only through the compiler-emitted `--hosted-tool-cache` runtime mode
- validate the selected cache root while preserving the existing fresh job-private cache when no image is configured
- document the opt-in contract and the Jammy toolchain image example

## Validation

- `mise run check`


sj26:

PR #103, opened 2026-08-10
Provide a default scoped GitHub token

lox:
## Why

Imported workflows currently receive a GitHub token only when they declare an explicit `permissions:` block. That differs from GitHub Actions and leaves actions that rely on an effective `github.token` default without credentials.

## What

Apply a narrow `contents: read` product default when permissions are omitted, while preserving explicit empty permissions as tokenless and limiting reusable workflows to the caller's effective authority. Token provisioning remains lazy, event-repository scoped, and redacted before use.


sj26:

PR #104, opened 2026-08-10
Support Node 16 actions with a deprecation warning

lox:
## Summary

- accept `runs.using: node16` and execute it with digest-verified managed Node 16.20.2
- emit GitHub Actions’ historical aggregate Node 16 deprecation warning as both a masked log warning and a Buildkite annotation
- name only invoked Node 16 actions, deduplicate repeated/pre/main/post invocations, and retain the warning across action failures, oversized-output suppression, and a saturated untrusted annotation budget
- preserve current GitHub-compatible `node20`/`node24` execution on managed Node 24.18.0
- document the compatibility boundary and cover host/container runtime selection

## Validation

- `DOCKER_CONTEXT=default mise run --jobs 1 check`
- focused Node 16 runtime tests under `go test -race`
- Oracle review; applied its annotation-priority finding before the full check

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,9 @@ for limits and failure behavior.
 As a quick screen, the plugin path is a good fit for workflows built from:
 
 - Linux Bash and `sh` steps;
-- JavaScript, composite, local, and anonymous public actions;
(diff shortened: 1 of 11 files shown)


sj26:

PR #107, opened 2026-08-10
Harden native checkout compatibility boundaries

lox:
## Why

The native `actions/checkout` adapter was selected by repository identity alone, so a moved major tag or arbitrary future commit could silently receive provider checkout authority without an exact-source compatibility decision. Its documented input/default and lifecycle differences were also too implicit.

## What

- admit only audited exact implementations for current [v4](https://github.com/actions/checkout/blob/11d5960a326750d5838078e36cf38b85af677262/action.yml), [v5](https://github.com/actions/checkout/blob/fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09/action.yml), [v6](https://github.com/actions/checkout/blob/d23441a48e516b6c34aea4fa41551a30e30af803/action.yml), [v7](https://github.com/actions/checkout/blob/3d3c42e5aac5ba805825da76410c181273ba90b1/action.yml), and the corpus's [v7.0.0-era pin](https://github.com/actions/checkout/blob/9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0/action.yml); unknown resolutions fail before source materialization where possible and again at compiler, credential preflight, and runtime boundaries
- make the bounded input contract explicit and tested, including empty/default values, boolean case variants, case-colliding names, exact event SHA output behavior, depth 0/1, tags, progress, anonymous and provider-credential paths, empty-workspace enforcement, and real temporary Git repositories
- document the full [v1](https://github.com/actions/checkout/blob/50fbc622fc4ef5163becd7fab6573eac35f8462e/action.yml) through [v7](https://github.com/actions/checkout/blob/3d3c42e5aac5ba805825da76410c181273ba90b1/action.yml) metadata/default/lifecycle evolution, native semantic differences, outputs, and credential boundary

Public-HTTPS submodules remain deliberately unsupported. A safe implementation would need complete direct/recursive `.gitmodules` URL and path validation, nested-manifest and symlink confinement, depth semantics, and proof that event-repository credentials cannot reach submodule hosts or paths. `true`, `recursive`, private, relative, and SSH submodules therefore all fail closed rather than receiving a corpus-specific implementation.

## Validation

- `git ls-remote https://github.com/actions/checkout.git 'refs/tags/v[1-7]' 'refs/tags/v[1-7]^{}'` — confirmed the documented exact v1-v7 targets (including peeled v1)
- `mise run check` — passed format, build, tests, race tests, Go/shell lint, vet, all 8 plan fixtures, all local smoke classifications, and GoReleaser config validation; `golangci-lint`: `0 issues`
- `mise run corpus:oss-profile` — expected non-zero mixed result without executing third-party code: **6 admitted, 4 blocked**; `jq-valgrind` now fails closed specifically on `submodules: true`

Tracking: [PB-2730](https://linear.app/buildkite/issue/PB-2730/compatibility-gap-expand-actionscheckout-input-and-repository). This PR is the bounded hardening slice; submodule, LFS, and alternate-repository compatibility remain open there.

### DIFF
--- a/docs/architecture/0003-protected-capability-control-plane.md
+++ b/docs/architecture/0003-protected-capability-control-plane.md
@@ -231,14 +231,17 @@ Content-Type: application/json
 }
 ```
 
-The compiler adds `provider-token-read` after resolving the exact
-`actions/checkout` adapter and validating its repository, ref, depth, path,
-submodule, and credential-persistence inputs. Fresh same-process compiler
-provenance permits that one capability through upload admission. The runtime
-uses `buildkite-agent git-credentials-helper` only when the immutable plan has
-that capability, a root checkout selector names the verified adapter, and the
-server-provided job environment indicates repository-provider Git credentials
-are enabled. Otherwise it performs the same checkout anonymously.
+The compiler adds `provider-token-read` only after the resolved
+`actions/checkout` commit is in the audited native-adapter allowlist and its
+repository, ref, depth, path, submodule, and credential-persistence inputs pass
+validation. Fresh same-process compiler provenance permits that one capability
+through upload admission. Before resolving Git or repository credentials, the
+runtime revalidates every checkout lock in the plan and requires a root checkout
+selector to name the verified adapter. It invokes
+`buildkite-agent git-credentials-helper` only when the immutable plan has that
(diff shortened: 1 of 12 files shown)


sj26:

PR #109, opened 2026-08-11
Avoid mise GitHub API rate limits in CI

lox:
## Why

Buildkite build 544 failed in the mise plugin's pre-command setup before repository checks could start. Cold hosted jobs had to discover pinned Aqua release assets through unauthenticated GitHub API requests, so the shared unauthenticated rate limit made CI availability depend on unrelated traffic. Injecting a token is not safe here because the pinned plugin leaves its environment available to potentially hostile PR commands.

## What

- commit a mise 2026.5.12-generated lockfile with direct URLs, SHA-256 checksums, and Aqua's verified provenance records for all supported platforms
- enable lockfile maintenance and make Buildkite consume it strictly, failing closed if a tool lacks a locked artifact instead of falling back to provider discovery
- document lock regeneration using the oldest mise version supported by CI

This keeps branch and fork PR checks credential-free while preserving pinned artifact checksum and provenance verification.


sj26:

PR #116, opened 2026-08-11
Replace the implementation plan with durable security guidance

lox:
## Why

The implementation plan had grown into a stale 18,000-word mix of completed history, future ideas, and release evidence. Current support belongs in the compatibility reference, while remaining work is tracked in Linear.

## What

- remove the retired plan rather than keeping a second source of truth
- add a concise operator-facing security model covering isolation, credentials, and untrusted inputs
- update documentation navigation and point historical evidence to Git, pull requests, Buildkite, and the smoke inventory


lox:
Acknowledged for review 4904841015: there is no concrete issue to address. The isolation and credential-boundary wording was separately checked against the runtime and canonical compatibility contract before publication.

sj26:

PR #117, opened 2026-08-11
Add hidden thin-plugin entrypoint

lox:
## Why

The GitHub Actions Buildkite plugin needs a deliberately narrow integration boundary instead of invoking the public `upload` command and translating plugin configuration into CLI arguments.

## What

Add a hidden, zero-argument `buildkite-gha plugin` entrypoint that reads the canonical workflow environment, normalizes symbolic Buildkite commits from checked-out `HEAD`, and delegates to the existing upload implementation. Keep the public command surface and direct `upload` behavior unchanged.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -97,13 +97,18 @@ so piping its YAML directly to `buildkite-agent pipeline upload` is incomplete.
 
 ## Upload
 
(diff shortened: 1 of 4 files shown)


sj26:

PR #118, opened 2026-08-11
Make compatibility status easier to scan

lox:
## Why

The compatibility reference is comprehensive, but its support states are easy to miss when scanning.

## What

- add a restrained five-icon status legend
- apply the same markers to the summary matrix, detailed status callouts, and support tables
- keep the text labels so meaning does not depend on emoji alone


sj26:

PR #124, opened 2026-08-11
Remove the superseded compatibility report API

lox:
## Why

The CLI's `validate`, `compile`, and `upload` commands emit only the stage-oriented processing report (`buildkite-gha/processing-report/v1`), but the older v1/v2 compatibility report machinery (`Report`, `ProfileReport`, their constructors, and `Write`/`WriteProfile`) still lived alongside it with no production caller. Several CLI tests also decoded the current processing-report JSON into those obsolete structs, which passed only because `encoding/json` ignores unknown fields — so they silently asserted against the wrong schema shape.

## What

- Delete the v1/v2 report types, constructors, writers, and their schema constants from `internal/compatibility`, plus their dedicated tests. The `Stage` type stays because the live `ProcessingReport` uses it for compile/admission results.
- Migrate the CLI tests that decoded into the obsolete structs to decode `ProcessingReport`, so command contract tests now check the real output schema.

`mise run check` passes.


sj26:

PR #125, opened 2026-08-11
Narrow unused internal API surface

lox:
Pre-1.0 cleanup: removes dead entry points and makes internal-only runtime identifiers private, shrinking the exported surface we have to keep stable without changing behavior.

- Delete `transport.Agent.UploadArtifact` — zero callers; `UploadArtifactFrom` / `UploadArtifacts` are the live paths.
- Delete the exported `DiscoverNode` / `DiscoverNode24` wrappers — production uses the private `discoverNodeContext` path; tests now exercise it directly.
- Unexport `JavaScriptAction`, `DockerAction`, `Runner.RunDocker` (now `runDockerAction`), and `VerifyWorkflow` — all callers are inside `internal/runtime`.

Kept out of scope: the `compiler.CompilePlans*` wrappers have no production callers but are the entry point for ~60 test call sites across `internal/compiler` and `internal/runtime`, so removing them would be a large test rewrite for no behavior benefit.

Validation: `mise run check` passes (format, build, test, race, lint, vet, plan-fixtures, smoke:local, release:check).

### DIFF
--- a/internal/runtime/actions.go
+++ b/internal/runtime/actions.go
@@ -167,7 +167,7 @@ func (r *actionLockResolver) verifyWorkspace(lock plan.ActionLock) (metadata.Met
 	if r.workspace == "" {
 		return metadata.Metadata{}, fmt.Errorf("workspace is missing")
 	}
-	if err := VerifyWorkflow(r.job, r.workspace); err != nil {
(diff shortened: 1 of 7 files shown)


sj26:

PR #133, opened 2026-08-12
docs: Update thin plugin guidance

lox:
## Summary

- replace the stale plugin/runtime version example with the current preview configuration
- document that the plugin delegates to the hidden `buildkite-gha plugin` entrypoint
- make the latest stable CLI the default and exact releases from 0.8.0 onward an optional pin
- align the security checklist with the same release-selection guidance

This documents the behavior introduced by buildkite-plugins/github-actions-buildkite-plugin#28 and should land with or after that plugin change.

## Validation

- `git diff --check`
- confirmed no `v0.4.4`, `v0.4.2`, or old released-plugin pin guidance remains in `README.md` or `docs/`

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ steps:
   - label: ":github: Test"
     key: "gha-ci"
     plugins:
-      - github-actions#v0.4.4:
+      - github-actions#main:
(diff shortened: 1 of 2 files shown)


sj26:

PR #134, opened 2026-08-12
Validate released plugin v0.8.0

lox:
## Why

The released-plugin demo still exercised github-actions v0.2.2, while v0.8.0 now resolves the latest buildkite-gha release through mise by default.

## What

- Run every released-plugin demo importer against github-actions v0.8.0 without an exact buildkite-gha version pin.
- Keep the demo contract test and development guide aligned with mise-based runtime resolution.
- Run the exact-source smoke outside explicit released-plugin demo builds so their identical generated job keys cannot collide; normal builds retain the exact-source smoke.


sj26:

PR #121, opened 2026-08-11
Remove obsolete compatibility experiments

lox:
## Why

The repository mixed current CI and hosted smoke coverage with dormant manual experiments, a superseded signed-plan prototype, and pre-release migration scaffolding. That made the supported validation paths harder to identify and left unused production-adjacent code behind.

## What

- Remove the shell/concurrent differential oracle, transport probe, signed-plan prototype, and migration POC infrastructure.
- Keep normal CI, behavior-specific hosted smoke, released-plugin validation, and production artifact/result transport.
- Rename the surviving cache fixture and example workflow identifiers around their durable purpose.
- Remove obsolete tasks, schemas, fixtures, tests, and documentation tied to the retired experiments.

PR #135, opened 2026-08-12
Remove the dead differential-smoke harness package

lox:
Pre-1.0 cleanup, continuing #124/#125.

- Delete `internal/harness` (803 lines): zero importers module-wide (`go list -deps ./...`), and no references from `scripts/`, CI, `mise.toml`, or docs. It served the differential smoke runners that #121 removed; only its own tests exercised it.
- Unexport `plan.MaxStepTargets` — used only inside `internal/plan`.
- Unexport `expression.ValidateRuntimeTemplate` — no production callers; kept (private) because its test cross-checks the runtime reference grammar against `Evaluate`.

`plan.MaxNeedOutputs` stays exported: `internal/compiler` uses it.

Validation: `mise run check` passes.


sj26:

PR #136, opened 2026-08-12
Concentrate the shared workflow-processing pipeline

lox:
First slice of concentrating processing-report construction (the report code has needed seven consecutive fix commits, largely because the same logic lives in three command flows).

`validate`, `compile`, and `upload` each hand-rolled the same pipeline: read workflow → emit environment-failure report → acquire event → emit event-failure report → validate → start the processing report → emit on rejection. `validate --profile` and `upload` additionally duplicated the hosted-preflight failure classification block almost verbatim.

This moves that pipeline into `internal/cli/processing_report.go` behind a small `processingOutput` value plus four helpers (`loadProcessingInputs`, `validatedProcessingReport`, `applyHostedPreflight`, `classifyHostedTokenlessFailure`). The three commands become thin callers; `cli.go` loses ~126 lines.

Intentional minor unifications (none pinned by tests):
- Report-write failures now always print one `write report` line to stderr; previously `validate` reported them, `upload` ignored them, and `compile` did both depending on the path.
- Removed an `if format == "ir-json"` branch in `compile` whose arms were identical.
- `upload` registers its signal context before reading the workflow file instead of after.

Exit codes and report contents are unchanged. A follow-up slice can move report semantics from `internal/cli` into `internal/compatibility` so the schema and its construction live together.

Validation: `mise run check` passes.

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -925,36 +925,17 @@ func validate(args []string, stdout, stderr io.Writer, version string) int {
 	if profile != "" && eventPath == "" {
 		return usageError(stderr, "validate: --event-path is required with --profile")
 	}
-	source, err := os.ReadFile(workflowPath)
-	if err != nil {
-		report := environmentProcessingReport(workflowPath, profile, "workflow input could not be read")
-		_ = compatibility.WriteProcessing(stdout, format, report)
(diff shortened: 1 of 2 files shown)


sj26:

PR #137, opened 2026-08-12
Move processing-report semantics into compatibility

lox:
Second slice of concentrating processing-report construction (follows #136).

The processing-report schema lives in `internal/compatibility` (checked against `schemas/processing-report-v1.schema.json`), but the semantics that populate it — stage transitions, diagnostic extraction from compiler findings, job/action result marking, evidence folding — lived in `internal/cli`. Report bugs therefore straddled two packages; the feature needed seven consecutive fix commits.

This moves that construction logic into `internal/compatibility/processing_compiler.go` (`compatibility` now imports `compiler`; no cycle, `compiler` does not import `compatibility`):

- `InitialProcessingReport`, `EnvironmentProcessingReport`, `EventInputProcessingReport` constructors
- `AddFailure`, `AddEnvironmentFailure`, `ApplyEvidence` methods on `ProcessingReport`
- private helpers (error flattening, diagnostic extraction, stage bookkeeping) move along and stay private

`internal/cli/processing_report.go` shrinks from 426 to 105 lines and now contains only command-facing IO adapters. The CLI-side stage-alias constants are gone; `cli.go` uses `string(compiler.Stage…)` at its three remaining sites and the test file keeps a local alias block for its report assertions.

Pure code motion plus renames; no behavior change. Validation: `mise run check` passes.

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -941,7 +941,7 @@ func validate(args []string, stdout, stderr io.Writer, version string) int {
 	if profile != "" {
 		_, _, distributionDigest, executableErr := executable()
 		if executableErr != nil {
-			addEnvironmentFailure(&processingReport, "compiler executable could not be inspected")
+			processingReport.AddEnvironmentFailure("compiler executable could not be inspected")
 			processingReport.Result = "indeterminate"
 			_ = out.write(processingReport)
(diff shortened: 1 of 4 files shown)


sj26:

PR #139, opened 2026-08-12
Remove legacy CompilePlans wrappers from the compiler

lox:
Part of the pre-1.0 architecture cleanup (follows #135, #136, #137).

`CompilePlans`, `CompilePlansWithOptions`, and `CompilePlansContext` had zero production callers — the CLI compiles through `CompileBundlePlansContext`. They survived only because ~66 test call sites used them, so tests exercised an interface that never ships, including a legacy `gha-untrusted` queue restriction users cannot reach.

## Changes

- Delete the three wrappers and the now-unused private `compilePlans` from `internal/compiler/compiler.go` (net −50 lines of production API).
- Migrate compiler and runtime tests to package-private helpers (`compilePlansForTest`, `compileUntrustedPlans`) that call the production `CompileBundlePlansContext` path and unwrap `plan.Job` from bundle artifacts. Tests now also run action-resolution validation, matching shipping behavior.
- Remove the wrapper-only "unsupported target queue" assertion from `TestDefaultCompilePolicyIsUntrustedAndUsesBuildkiteDefault`; it validated input checking on the deleted convenience wrapper, not compiler queue policy. Queue admission remains covered by the options-based policy tests.

## Validation

`mise run check` passes (format, build, test, race, lint, vet, local smoke, release check).

### DIFF
--- a/internal/compiler/actions_test.go
+++ b/internal/compiler/actions_test.go
@@ -421,7 +421,7 @@ runs:
 		if err := os.WriteFile(workflowPath, []byte(workflow), 0o644); err != nil {
 			t.Fatal(err)
 		}
-		return CompilePlansWithOptions(workflowPath, []byte(workflow), pushEvent(t), "0.0.0-test", testDistributionDigest, defaultOptions())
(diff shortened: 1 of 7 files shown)


sj26:

PR #140, opened 2026-08-12
Introduce invocationEnvironment for step env overlay and PATH precedence

lox:
Part of the pre-1.0 architecture cleanup (follows #135, #136, #137, #139). First bounded slice of the runtime environment work — no package split, no `Runner` rewrite.

`runActionStep` computed `mergeStepEnvironment(jobEnv, stepEnv)` four times and inlined a dense Docker PATH-precedence block reading three booleans from two maps and two `Runner` fields. The overlay order and PATH rules were order-sensitive and unnamed.

## Changes

- New `invocationEnvironment` value, constructed once at the top of `runActionStep`, owns:
  - `process()` — the environment for shell and JavaScript step processes (job env, then step env, runtime context names protected).
  - `docker()` — overlays action-declared env beneath the invocation env; the action may replace `PATH` only when neither the job nor the step pinned one, and it reports whether the Docker image `PATH` must yield.
- The Docker branch's inline PATH block is a direct transcription into `docker()`; behavior is unchanged.
- New focused unit tests cover overlay order, runtime-context protection, action/input overlay semantics, and every PATH precedence combination (implicit job PATH, explicit job PATH, step PATH, action PATH).

Composite and remote-preparation paths are deliberately untouched; they can adopt the same value in a later slice if it pays for itself.

## Validation

`mise run check` passes (format, build, test, race, lint, vet, local smoke, release check).

### DIFF
--- a/internal/runtime/invocation_env_test.go
+++ b/internal/runtime/invocation_env_test.go
@@ -1,0 +1,102 @@
+package runtime
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestInvocationEnvironmentProcessOverlayOrder(t *testing.T) {
+	var r Runner
+	environment := r.invocationEnvironment(
+		map[string]string{"SHARED": "job", "JOB_ONLY": "job", "GITHUB_REPOSITORY": "owner/repo"},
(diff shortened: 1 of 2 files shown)


sj26:

PR #141, opened 2026-08-12
Keep the hosted toolchain image current with Renovate

lox:
## Why

The hosted toolchain runtime image is digest-pinned, but the digest-only Namespace reference does not tell Renovate which agent-base variant to follow.

## What

Add a focused Renovate regex manager that tracks the `ubuntu-noble-hosted-toolchains` Docker tag and updates the immutable Namespace digest. Refresh the documented reference to the digest published by the latest agent-base-images build.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -128,7 +128,8 @@ Set an image by digest:
 
 ```yaml
 env:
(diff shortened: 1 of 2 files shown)


sj26:

PR #142, opened 2026-08-12
Clarify GitHub token compatibility boundaries

lox:
## Why

The compatibility contract names the token policy but does not explain that a local reusable-workflow call blocks token issuance for every job in the selected workflow. This also makes `actions/checkout` request a token through its default `github.token` input.

## What

Make the reusable-workflow boundary visible in the support matrix and detailed token reference. Keep the merge-queue and GitHub Enterprise Server exclusions explicit.

Refs #160.


sj26:

PR #145, opened 2026-08-12
Support bounded multi-prefix artifact selection

lox:
## Why

Some workflows select artifacts from a small fixed set of producer prefixes, but the native download adapter currently accepts only one bounded glob.

PostHog uses the same two-prefix pattern in its [`django_tests` failure gate](https://github.com/buildkite/posthog/blob/b9a0729f733e6dd170017677bb339608489ec3d2/.github/workflows/ci-backend.yml#L3463-L3471) and [`report-test-timings`](https://github.com/buildkite/posthog/blob/b9a0729f733e6dd170017677bb339608489ec3d2/.github/workflows/ci-backend.yml#L3809-L3815):

```yaml
- name: Download junit artifacts
  uses: actions/download-artifact@634f93cb2916e3fdff6788551b99b062d0335ce0
  with:
    path: ./junit-artifacts
    pattern: '{junit-results-backend,product-junit-results}-*'
```

Previously the braces caused admission to reject the pattern. This PR expands that exact bounded form into:

```text
junit-results-backend-*
product-junit-results-*
```

and evaluates both against the same verified artifact inventory.

## What

- Accept one optional leading group of 2–8 comma-separated literal prefixes.
- Reject empty alternatives, nesting, braces elsewhere, and glob metacharacters inside alternatives.
- Validate every expanded glob against the existing 255-byte artifact-name pattern contract.
- Preserve verified direct-`needs` provenance, the 64-match limit, deterministic artifact-name ordering, and overlap deduplication.
- Add PostHog-shaped compiler coverage plus runtime order, deduplication, collision, match-limit, staging, and failure-path tests.

This fixes bounded multi-prefix parsing and matching only. The unchanged PostHog uses still have independent blockers—most notably omitted `merge-multiple: true`, plus destination/provenance constraints in their respective jobs—so brace alternation remains outside the advertised public compatibility contract until hosted multi-producer proof exists.

Related: [PB-2796](https://linear.app/buildkite/issue/PB-2796/support-bounded-multi-prefix-artifact-selection)

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -557,15 +557,15 @@ The adapter sets `artifact-id` and `artifact-digest`. The `artifact-url` output 
 | Input | Supported values |
 | --- | --- |
 | `name` | Exact name, mutually exclusive with `pattern`; runtime expressions are allowed. |
-| `pattern` | Bounded artifact-name glob or leading literal-prefix alternation; requires `merge-multiple: true`; runtime expressions are allowed. |
+| `pattern` | Bounded artifact-name glob; requires `merge-multiple: true`; runtime expressions are allowed. |
 | `path` | Optional literal workspace-relative path. |
 | `merge-multiple` | Omitted or `false` with `name`; required `true` with `pattern`. |
 | v8 `skip-decompress` | Omitted or `false`. |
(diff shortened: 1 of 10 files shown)


sj26:

PR #150, opened 2026-08-12
Use portable rooted reads for artifact uploads

lox:
## Why

The native upload-artifact adapter uses `unix.Openat`, which is unavailable when compiling for Darwin. This prevents the runtime from building for supported native macOS jobs.

## What

Replace the handwritten descriptor traversal with the standard library's portable `os.Root`. Artifact reads remain confined to the resolved workspace, reject symlink components before opening, use nonblocking opens for special-file replacement, and retain the existing source identity and mutation checks.

Extend replacement coverage to both file and intermediate-directory symlinks.

### DIFF
--- a/internal/runtime/upload_artifact.go
+++ b/internal/runtime/upload_artifact.go
@@ -17,12 +17,12 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"syscall"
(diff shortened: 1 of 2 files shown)


sj26:

PR #151, opened 2026-08-12
Document the strict and loose expression evaluator families

lox:
First of three expression-package cleanups reviewed in the pre-1.0 architecture thread (follows #135–#140). No behavior change.

The package has two deliberately different value-semantics families: `condition*` (strict — mixed-type equality fails closed, used by runtime conditions) and `actionInputDefault*` (GitHub's loose coercion for action input defaults). Nothing recorded that the split is intentional, which invites an accidental "deduplication" into a bug.

## Changes

- Doc comments on both families stating the boundary and that they must not be unified; `conditionNumber` marked as shared conversion infrastructure that does not itself imply coercion.
- Characterization tests for `compileInputLiteral`, the least-covered function in the package (37.5%): null/bool/string/number representations, apostrophe doubling, and fail-closed rejection of aggregates and non-YAML numeric types.

## Validation

`mise run check` passes.

### DIFF
--- a/internal/expression/expression.go
+++ b/internal/expression/expression.go
@@ -1087,6 +1087,12 @@ func resolveConditionReference(root string, path []string, context ConditionCont
 	return nil, fmt.Errorf("condition references unavailable value %s.%s", root, strings.Join(path, "."))
 }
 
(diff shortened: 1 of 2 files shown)


sj26:

PR #152, opened 2026-08-12
Split expression package into files by semantic concern

lox:
Second of three expression-package cleanups (follows #151). Pure code motion — no API, naming, or behavior change, and `expression_test.go` is deliberately untouched so the unchanged tests prove the move preserved behavior.

`expression.go` was a single 1790-line file containing several distinct semantic surfaces. Every declaration now moves verbatim (with its doc comment) into a file per concern, verified by line-multiset comparison against the original.

| File | Lines | Owns |
|---|---|---|
| `expression.go` | 251 | package doc, owned source types, parsing front doors, cross-concern helpers |
| `compile.go` | 345 | `CompileContext`, graph-time substitution/evaluation |
| `condition.go` | 531 | `ConditionContext`/`ConditionScope`, validation, strict fail-closed condition evaluation |
| `action_input_default.go` | 240 | the loose-coercion action input default family |
| `runtime.go` | 261 | direct runtime template evaluation and reference resolution |
| `references.go` | 216 | the narrow reference-introspection predicates |

Keeping `action_input_default.go` separate from `condition.go` preserves the documented strict-vs-loose evaluator boundary (#151).

Noted while splitting (not changed here, motion only): `validateRuntimeTemplate` has no production callers, only tests.

## Validation

`mise run check` passes.


sj26:

PR #153, opened 2026-08-12
Move action lifecycle condition evaluation into expression

lox:
Third of three expression-package cleanups (follows #151, #152), closing the one real encapsulation leak found in the review.

`evaluateLifecycleCondition` in `internal/runtime/job.go` hand-stripped `${{ }}` and reimplemented `success()`/`failure()`/`cancelled()`/`always()` with a switch — a second, private expression evaluator living outside the expression package.

## Changes

- The function moves **verbatim** to `expression.EvaluateActionLifecycleCondition` in `condition.go`. It deliberately does not delegate to `EvaluateCondition`: the lifecycle grammar is narrower and its semantics are exceptional — empty means unconditionally true (general conditions apply the implicit success guard), `failure()` means unsuccessful **and not cancelled** (cancellation dominates), and only a single bare status function with optional delimiters is accepted; anything else fails closed.
- The narrow name is intentional, so callers don't assume the broader condition grammar.
- Table tests pin the full truth table: empty/whitespace at success, failure, and cancellation; all four status functions with and without delimiters; case and whitespace handling; and fail-closed rejection of literals, references, compound expressions, arguments, unknown functions, and partial delimiters.
- All six runtime call sites adopt it unchanged — validation timing and error wrapping are preserved, including the deliberately ignored second evaluation error after prior validation.

## Validation

`mise run check` passes.

### DIFF
--- a/internal/expression/condition.go
+++ b/internal/expression/condition.go
@@ -321,6 +321,32 @@ func validateConditionReference(root string, path []string, scope ConditionScope
 	}
 }
 
+// EvaluateActionLifecycleCondition evaluates an action pre-if or post-if
+// condition against the supplied lifecycle state. The accepted grammar is
+// deliberately narrower than general conditions: exactly one bare status
+// function with optional ${{ }} delimiters. An empty condition is
+// unconditionally true (unlike general conditions, which apply the implicit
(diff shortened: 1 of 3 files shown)


sj26:

PR #154, opened 2026-08-12
feat: add GitHub workflow access tokens

wolfeidau:
## Why

GitHub Actions workflows can depend on `github.token` or `secrets.GITHUB_TOKEN`. `buildkite-gha` needs a secure way to provide that token without using ambient credentials or broadening access beyond the compiled workflow policy.

## What

- Mint short-lived, repository-scoped GitHub tokens for eligible workflow jobs.
- Keep action source authentication on the existing scoped-token path.
- Document the supported permission and workflow constraints.

## How

The compiler records verified policy evidence from the workflow path and explicit top-level permissions. Hosted admission rejects workflows with missing evidence, job-level permission overrides, reusable workflow jobs, invalid filenames, or unsupported permissions. At runtime, the CLI sends the repository, workflow filename, and effective permissions to the workflow-token endpoint, then registers the returned token for log redaction.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -149,9 +149,9 @@ permissions:
   pull-requests: write
 ```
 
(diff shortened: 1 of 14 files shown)


lox (approved):

tehpeh (approved):
Looks good, Cursor and Grok agree.

sj26:

PR #155, opened 2026-08-12
Run the full Go test suite on macOS arm64

lox:
## Why

The Darwin cross-build catches compilation errors but cannot execute macOS binaries or verify Darwin filesystem and process behavior. Native macOS support needs a runtime regression guard.

## What

Add a required Buildkite step on the `mac-small` queue that verifies the agent is Darwin/arm64, builds every Go package, and runs the ordinary uncached `go test ./...` suite.

Mark Linux/amd64 importer and Docker tests with Go build constraints or focused `t.Skip` guards. Portable tests still run on both platforms. Canonicalize temporary paths in alias-sensitive fixtures so macOS's `/tmp` alias does not produce false path-escape failures.

Keep every other checked-in pipeline step pinned to the hosted queue.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -104,7 +104,7 @@ steps:
           version: "2026.5.12"
     command: "mise run --jobs 1 check"
 
-  - label: ":go: macOS arm64"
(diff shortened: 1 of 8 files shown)


sj26:

PR #156, opened 2026-08-12
Clarify v0.9.0 production availability

lox:
## Why

Runtime v0.9.0 includes native macOS and runner context support, but released plugin v0.8.0 cannot declare runner mappings. The documentation currently presents the runtime capability as production plugin availability.

## What

Clarify that Linux `runner.os` and `runner.arch` work through plugin v0.8.0 with runtime v0.9.0, while native macOS remains runtime-only until the companion plugin release. Document the supported runner fields, values, and expression boundaries, and use released versions in the production Linux examples.


sj26:

PR #158, opened 2026-08-12
Fix remaining native macOS portability gaps

lox:
## Why

#150 made native artifact uploads portable, but the action/runtime path still had Darwin build blockers and rejected valid macOS paths whose ancestors resolve through aliases such as `/var` to `/private/var`. Maintainer proof scripts also assumed Bash 4 and GNU checksum tools.

## What

Replace the remaining download-artifact descriptor syscalls with confined `os.Root` operations while preserving staging identity, collision, special-file, and atomic publication checks. Canonicalize trusted action and mise roots so aliased ancestors work without accepting symlinks at or below those roots.

Move Unix-only flags and test operations behind build tags, and make maintainer scripts compatible with macOS Bash 3 and `shasum -a 256`.


lox:
@buildsworth-bk review

sj26:

Issue #160, opened 2026-08-12
Support GitHub tokens across local reusable workflows


lox:
> This tracks a planned compatibility outcome. The [compatibility contract](https://github.com/buildkite/buildkite-gha/blob/main/docs/compatibility.md#github-token) describes current support; an open issue is not current support or an ETA.

## Workflow

Caller:

```yaml
on: push
permissions:
  contents: read
jobs:
  quality:
    uses: ./.github/workflows/quality.yml
```

Called workflow:

```yaml
on: workflow_call
jobs:
  quality:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v7
      - run: ./quality.sh
```

## Current behavior

Compilation detects the checkout metadata default `token: ${{ github.token }}` and requests a bounded job token. Hosted issuance then rejects the request because the caller contains a reusable-workflow job.

The same tokenless reusable-workflow composition is supported.

## Expected outcome

Issue a job-bound `GITHUB_TOKEN` to eligible direct and expanded jobs across local reusable-workflow boundaries without expanding repository or permission authority.

## Done when

- Static `secrets.GITHUB_TOKEN` references work in eligible jobs expanded from local reusable workflows.
- Effective action input defaults that reference `github.token`, including supported `actions/checkout` releases, work in eligible expanded jobs.
- Caller and callee permissions only narrow the token's compiler-verified scope.
- The existing event-repository, provenance, feature-setting, fork, merge-queue, and server-side policy boundaries still apply.
- Direct jobs do not lose token eligibility merely because another job calls a local reusable workflow.
- Remote and private reusable workflows remain unsupported.
- Hosted proof covers direct, one-level, nested, and denied permission cases.
- The compatibility contract documents the shipped behavior before this issue closes.

## Evidence

- [Buildkite rejected the reusable-workflow token request](https://buildkite.com/buildkite/buildkite-gha-examples/builds/45).
- [GitHub Actions passed the same commit](https://github.com/buildkite/buildkite-gha/actions/runs/31575531200).

lox:
The deliberately limited reusable-workflow token MVP has now shipped:

- buildkite/buildkite#32609 accepts immutable top-level workflow policies containing reusable calls and continues enforcing repository, build provenance, feature/pipeline opt-in, pull-request, merge-queue, and server-policy boundaries.
[...]

PR #161, opened 2026-08-12
Update tokenless workflow examples

lox:
## Why

The advanced comparison example requested `github.token` through checkout inside a reusable workflow, which is outside the current hosted compatibility boundary and caused the Buildkite side to fail. The examples importer also used the retired source configuration from plugin v0.4.4.

Related to #160.

## What

Keep the advanced example within the tokenless reusable-workflow subset by demonstrating static package input and output propagation without checkout. Update both examples importers to plugin v0.9.3 and its `source-ref` configuration.


sj26:
Released in [v0.10.2](https://github.com/buildkite/buildkite-gha/releases/tag/v0.10.2)

PR #162, opened 2026-08-12
Isolate fallback event tests from tag builds

lox:
## Why

The v0.10.0 tag build inherited `BUILDKITE_TAG` while testing a synthetic pull request, producing contradictory event inputs. Main and local checks did not expose the tag-only environment leak.

## What

Clear the tag in the fallback-event test before constructing each synthetic Buildkite event.


sj26:

PR #165, opened 2026-08-12
Align push webhook refs for PR-associated builds

lox:
## Why

Buildkite can associate a webhook-created push build with an open pull request. The webhook then selected `push` while the compatibility snapshot retained `refs/pull/<n>/head`, causing trigger translation to reject otherwise valid plugin uploads.

## What

When linked webhook metadata selects `push`, derive the authoritative ref from the Buildkite tag or branch. Keep the existing pull-request ref behavior for pull-request events and cover the mixed push/PR-association case with a regression test.


sj26:
Released in [v0.10.3](https://github.com/buildkite/buildkite-gha/releases/tag/v0.10.3)

PR #166, opened 2026-08-12
Fix flaky service cleanup test

lox:
## Why

The service cleanup failure test accepted any container startup error as its intended second-service create failure. Under parallel subprocess load, an earlier setup failure could therefore be misreported as incomplete cleanup.

## What

Run this subprocess-heavy failure-injection test serially and require the expected second-service create error before checking exact cleanup calls.


sj26:
Released in [v0.11.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.11.0)

PR #169, opened 2026-08-12
Name the policy rule that rejects a runs-on target

JuanitoFatas:
Every rejected `runs-on` produced the same sentence, so a Windows row and an
unmapped macOS row looked identical. Only the first is a real boundary. The
second just needs a queue in the plugin's `runners:` config.

`ProcessingFinding.Message` hides the detailed error on purpose: `runs-on` can
interpolate `${{ github.event.* }}`, so a resolved label is untrusted. The
message now appends the rejection reason only. Reasons are literals from a
fixed set, so no resolved value can reach a report.

Before:

    resolved runner target is not admitted by policy
    resolved runner target is not admitted by policy

After:

    resolved runner target is not admitted by policy: unsupported operating system
    resolved runner target is not admitted by policy: runner label is not mapped by policy

Detailed error text is unchanged, so the Go API and its existing tests see
exactly what they saw before.

<details>
<summary>How we found this improvement</summary>

Not by reading the code. By running `validate` against unmodified workflows from
eight public repos — bubbletea, execa, commander.js, ripgrep, black, cli/cli,
ruff, and p-queue — and comparing what each diagnostic actually told us.

Most were already actionable: `condition function "startsWith" is unsupported`,
`reusable workflow "…@main" is remote or runtime-dependent`. The runner
rejection was the outlier — the most frequent diagnostic in the sample, and the
only one that named nothing. On ripgrep it printed four identical lines,
separable only by an instance hash. `RunnerPolicy.resolve` had known the exact
cause all along.

The first attempt was to drop the `Message` override and let the detailed error
through. That was wrong, and the test name caught it:
`TestProcessingReportRedactsEventDerivedRunnerValues`. `runs-on` can be
`${{ github.event.runner }}`, so a resolved label may carry event payload data.
The override is a redaction control — that test plants a sentinel in the payload
and fails if it reaches stdout or stderr.

That reframed the change from "show the error" to **report why, never what**,
which is what this PR does.

</details>

### DIFF
--- a/internal/cli/cli_test.go
+++ b/internal/cli/cli_test.go
@@ -793,7 +793,7 @@ func TestRunValidateAndCompile(t *testing.T) {
 		if err := json.Unmarshal(stdout.Bytes(), &report); err != nil {
 			t.Fatal(err)
 		}
-		if report.Result != "incompatible" || report.Compile.Result != "incompatible" || len(report.Diagnostics) != 1 || report.Diagnostics[0].Message != "resolved runner target is not admitted by policy" {
+		if report.Result != "incompatible" || report.Compile.Result != "incompatible" || len(report.Diagnostics) != 1 || report.Diagnostics[0].Message != "resolved runner target is not admitted by policy: runner label is not mapped by policy" {
 			t.Fatalf("macOS profile report = %#v", report)
 		}
 	})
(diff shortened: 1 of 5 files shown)


sj26:
Released in [v0.11.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.11.0)

PR #170, opened 2026-08-12
Skip unsupported workflows in wildcard uploads

dannymidnight:
## Summary

- treat the exact `workflows: *` selector as discovery mode
- report and omit incompatible or non-admitted workflows while uploading supported workflows
- keep explicit selectors and shared event, environment, action-resolution, artifact, and upload failures strict
- succeed without adding steps when every discovered workflow is unsupported

## Testing

- `mise run check`


dannymidnight:
We're think we might bail on the wildcard behaviour and instead be explicit so closing this one.

PR #182, opened 2026-08-12
Fix unused testFileLock and testExec lint findings

JuanitoFatas:
## Summary

`make lint` failed on darwin with two `unused` findings in
`test_platform_unix_test.go`: `testFileLock` and `testExec`.

## Root cause

Both helpers are called only from `containers_test.go`, which is
build-tagged `linux && amd64`. `test_platform_unix_test.go` tags them
`linux || darwin`, a broader set. On darwin (and linux/arm64), the
caller is excluded from the build, so the two symbols become dead
code and `unused` flags them.

## Fix

This is the "used only under another build tag" case. Moved
`testFileLock` and `testExec` into `containers_test.go`, whose build
tag already matches where they're actually used. No deletion, no
`nolint`, no config changes.

## Verified

- `make lint` — 0 issues (was 2 unused findings)
- `go build ./...` — passes
- `go test ./internal/runtime/...` — passes
- `GOOS=linux GOARCH=amd64 go vet ./internal/runtime/...` — passes (compile-checks the moved code under its real target tag)
- `GOOS=windows go build ./...` — passes
- `GOOS=darwin go build ./...` — passes

### DIFF
--- a/internal/runtime/containers_test.go
+++ b/internal/runtime/containers_test.go
@@ -17,6 +17,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"syscall"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


sj26:

PR #183, opened 2026-08-12
Admit implicit read-only workflow tokens

lox:
## Why

The compiler already resolves omitted workflow permissions to `contents: read`, but hosted admission rejected the corresponding workflow-policy evidence. Token-requiring actions could therefore fail admission despite requesting only the deterministic read-only default.

## What

- Accept omitted top-level permissions as exactly `contents: read`, without inheriting GitHub repository or organization settings.
- Preserve no-token behavior for explicit empty/`none` permissions and hosted denials for job-level maps and reusable-workflow jobs; writes still require an explicit top-level map.
- Cover compiler, bundle, checkout, and hosted CLI boundaries and document the immutable workflow-token endpoint contract.

The paired `buildkite/buildkite` backend PR is being developed separately.


lox:
Acknowledged review 4916884880. There are no actionable implementation findings to address; this remains queued for human security review. The failed Buildkite check was an unrelated golangci-lint cache/path issue referencing a nonexistent checkout from another agent, so I retried that job.

sj26:

PR #184, opened 2026-08-12
Move the lint fan-out from Makefile to mise.toml

JuanitoFatas:
`lint` was the only Makefile target with logic of its own — it listed `lint:go` and `lint:shell`, and `check` listed the same pair again. Now `mise run lint` owns that pair, and `check` and the Makefile both call it.

CI already runs `mise run --jobs 1 check` (`.buildkite/pipeline.yml`), so no pipeline change was needed.

Checked against mise 2026.5.12, the version the pipeline pins: a depends-only task returns a failed dependency's exit code, so a lint failure still fails the step. Full `mise run --jobs 1 check` passes.


lox (approved):
Thanks!

sj26:

PR #186, opened 2026-08-12
Use dedicated GitHub action source tokens

lox:
## Why

Buildkite's dedicated GitHub action-source token API replaces the generic scoped-token endpoint and limits issuance to 10 attempts per importer job per hour. Per-resolver minting can exceed that budget when one upload compiles many workflow roots.

## What

- Use the repository-only `github_action_source_access_token` endpoint without caller-supplied permissions or an old-endpoint fallback.
- Lazily mint and redact one token per importer job, then reuse it across workflow roots and nested composite action resolution.
- Keep disabled or unavailable rollout states on the safe anonymous path while preserving public-only visibility checks, anonymous codeload, digest verification, and exact-SHA bypasses.


sj26:

PR #188, opened 2026-08-12
Accept action metadata used by starter workflows

lox:
## Why

The hosted compatibility profile rejects the Maven starter workflow because current public actions declare inert metadata fields that the strict action parser does not recognize.

## What

Accept top-level `deprecationMessage` and action input `type` metadata while retaining strict parsing for unknown fields. Cover both fields in the action metadata tests.


sj26:
Released in [v0.12.1](https://github.com/buildkite/buildkite-gha/releases/tag/v0.12.1)

PR #190, opened 2026-08-12
Preserve values in compatibility annotation code spans

lox:
## Why

Follow-up to #187: annotation code spans pre-escaped `&`, `<`, and `>`, so Buildkite displayed valid workflow values such as `./.github/actions/a&b` as HTML entities.

## What

Leave code-span content unescaped for the Markdown renderer while retaining the variable-length backtick delimiter. The focused test covers HTML and Markdown metacharacters together.


sj26:

PR #191, opened 2026-08-12
docs: bypass mise release age for CLI install

wolfeidau:
Update the CLI installation command to bypass mise's default release-age delay. This prevents macOS installs from resolving to an older Linux-only release during the delay window.


sj26:

PR #193, opened 2026-08-13
Support native macOS arm64 importers

lox:
## Why

The released CLI already includes a Darwin arm64 artifact, but the importer rejects that platform and assumes its executable is the Linux job runtime. This prevents the plugin importer from running natively on macOS and makes importer targeting easy to confuse with generated-job queues.

## What

Allow importer execution on Linux amd64 or Darwin arm64. Plan applicable workflows first, then acquire, verify, and upload only the required same-release runtime distributions. Reuse the importer executable only for its matching platform and fail before artifact upload when a required counterpart is unavailable. Clarify the independent importer and generated-job targeting contracts in the docs.

Companion plugin change: buildkite-plugins/github-actions-buildkite-plugin#44.


lox:
Acknowledged review 4923121133. No code changes requested; the cross-platform runtime selection and pre-upload failure paths were covered by targeted tests, the full local check, and an independent Oracle review.

sj26:

PR #196, opened 2026-08-13
Style all Buildkite annotations

mitchbne:
## Description

PR #144 introduced compact headings and gray borders for workflow-command annotations, but processing diagnostics, job summaries, and corpus annotations still used unstyled output.

Apply the same treatment to every annotation created by buildkite-gha. Diagnostic content is HTML-escaped, oversized annotations preserve balanced markup, and authored job-summary Markdown remains unchanged beneath the styled heading.

## Preview

Styled workflow diagnostics annotation

## Deployment

Low risk. This changes annotation presentation only.

## Rollback

Revert this PR to restore the previous annotation formatting.

@buildsworth-bk approve up to L2

PR #200, opened 2026-08-13
Improve skipped and failed workflow replacement steps

dannymidnight:
## Summary

- Emit ignored workflow replacements as command steps without a no-op `command: ":"`.
- Move applicable workflow compilation diagnostics into the failed replacement job as job-scoped annotations.
- Format annotations with an h4 heading, a bold primary error, and a repository-relative workflow path in inline code.
- Keep fatal diagnostics that cannot create a replacement job on the importer.

## Testing

- `mise run check`
- `mise exec -- go test ./internal/cli`


dannymidnight:
eg.

CleanShot 2026-08-13 at 22 32 12@2x

internal/buildkite/pipeline.go
```diff
@@ -238,7 +238,7 @@ func emitWorkflow(out *bytes.Buffer, pipeline Pipeline, workflow preparedWorkflo
 		if workflow.Condition != "" {
 			_, _ = fmt.Fprintf(out, "    if: %s\n", yamlScalar(workflow.Condition))
 		}
-		command := `printf '%s\n' ` + shellQuote(failure.Message) + ` && exit 1`
+		command := `buildkite-agent annotate --scope=job --style=error ` + shellQuote(failure.Annotation) + ` && exit 1`
```
mitchbne: Can we do both? I.e. print in the log _and_ make an annotation?

mitchbne (commented):
I think we should call the object "Message" not "Annotation", but we should annotate and print the message

internal/buildkite/pipeline.go
dannymidnight: oh true, yeah we can do that.

internal/cli/processing_report.go
```diff
@@ -88,14 +89,15 @@ func processingAnnotation(report compatibility.ProcessingReport) (style, body st
 	}
 
 	var out strings.Builder
-	out.WriteString("### GitHub Actions workflow diagnostics\n\n")
-	out.WriteString("**Workflow:** ")
-	out.WriteString(markdownText(report.Workflow))
+	out.WriteString("#### GitHub Actions workflow diagnostics\n\n")
+	out.WriteString("**Workflow**: ")
+	out.WriteString(markdownCode(processingAnnotationWorkflowPath(report.Workflow)))
```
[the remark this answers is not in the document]
mitchbne: Could we also rebase this off https://github.com/buildkite/buildkite-gha/pull/196 so that we keep the styling aligned

internal/buildkite/pipeline.go
dannymidnight: Updated in f88399b: the failure command prints the original plain-text diagnostics first, then streams the annotation body over stdin to the job-scoped annotation command before exiting 1.

mitchbne: Should we print first just in case they annotation api is down or we have an argument error in the annotation creation

internal/cli/processing_report.go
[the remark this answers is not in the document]
dannymidnight: yessss! I thought I must have imagined a PR kicking around for styling.

internal/buildkite/pipeline.go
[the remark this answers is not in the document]
dannymidnight: Addressed in f88399b. The plain-text diagnostics are printed first, then the annotation body is piped to `buildkite-agent annotate` over stdin so it is not passed as a large argv element.

dannymidnight: Resolved in https://github.com/buildkite/buildkite-gha/pull/200/changes/f88399b68659ceb05fa65522d0cb0985d5ed08b3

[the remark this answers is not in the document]
dannymidnight: Resolved in https://github.com/buildkite/buildkite-gha/pull/200/changes/f88399b68659ceb05fa65522d0cb0985d5ed08b3

dannymidnight:
OK I think that's pretty good now. We've got job-scoped annotations, the command is more readable (so hopefully less confusing), and the logs are coloured so the error stands out.

CleanShot 2026-08-14 at 07 23 31@2x

CleanShot 2026-08-14 at 07 24 57@2x

sj26:
Released in [v0.13.1](https://github.com/buildkite/buildkite-gha/releases/tag/v0.13.1)

PR #201, opened 2026-08-13
Fix Origin action token guard evaluation

dannymidnight:
## Summary

- expose the event repository provider through `github.server_url` and `GITHUB_SERVER_URL`
- evaluate provider-only action input guards before requesting `GITHUB_TOKEN`
- avoid adding a GitHub workflow token when a non-GitHub provider cannot reach the token branch
- keep direct and dynamically guarded token references fail-closed

## Context

Public actions can define provider-sensitive optional token defaults:

```yaml
inputs:
  token:
    default: ${{ github.server_url == 'https://github.com' && github.token || '' }}
```

The compiler previously treated any syntactic `github.token` reference as unconditional. For a non-GitHub event repository, that added a GitHub workflow token to the job plan and then failed validation with `GitHub workflow token requires a valid github.com event repository`.

This change evaluates guards containing only `github.server_url` and `github.token`. If another runtime value affects reachability, compilation continues to require the token and fails closed.

## Verification

- `mise run check`
- generic regression coverage for a public action with a GitHub-only conditional token default

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -457,7 +457,7 @@ Job and step `if` conditions support literals and the syntax listed above. Order
 
 | Context | Job `if` | Step `if` |
 | --- | --- | --- |
-| `github.actor`, `github.event_name`, `github.ref`, `github.repository`, `github.server_url`, `github.sha` | ✅ Yes | ✅ Yes |
+| `github.actor`, `github.event_name`, `github.ref`, `github.repository`, `github.sha` | ✅ Yes | ✅ Yes |
 | `github.head_ref`, `github.ref_name` | ❌ No | ❌ No |
 | `runner.os`, `runner.arch` | ✅ Yes | ✅ Yes |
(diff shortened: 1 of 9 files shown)


dannymidnight:
Noice. Confirmed it works via `source-ref`.

wolfeidau:
@dannymidnight Found one issue.

- [P2] github.server_url is documented but rejected in conditions — docs/compatibility.md:460 marks it supported for job and step if, but
  internal/expression/condition.go:309 omits it from the allowed properties. A condition such as if: github.server_url == 'https://origin.cursor.com' fails validation. Add
  it to condition validation and compileContext for conditions combined with github.event, or remove the documentation claim.

Validation:

- Fetched origin/fix/origin-conditional-github-token at 59c2021.
- Build, tests, race tests, lint, vet, smoke tests, and release configuration checks passed.
- Working tree remains clean on main; local main is one commit behind fetched origin/main.

That will be $10.

sj26:
Released in [v0.13.2](https://github.com/buildkite/buildkite-gha/releases/tag/v0.13.2)

PR #203, opened 2026-08-13
Run workflow diagnostic steps unconditionally

wolfeidau:
## Summary

- keep generated diagnostic replacement steps unconditional
- retain translated trigger conditions on actual workflow groups
- cover pull request base-branch selection separately from push branch filtering

## Root cause

`failedGeneratedWorkflow` copied the workflow's translated trigger condition onto the top-level replacement step that downloads and publishes diagnostic artifacts. When the trigger condition was false—for example, a `chore_updates` push against `push.branches: [main]`—Buildkite skipped the diagnostic command, so the annotation explaining the filtered workflow never appeared.

Failure replacement steps now omit the trigger condition, and aggregate pipeline validation permits those steps without a condition or skip reason. Normal generated workflow groups continue to use their translated trigger conditions.

## Tests

- `mise exec -- go test ./internal/buildkite ./internal/cli -count=1`

### DIFF
--- a/internal/buildkite/pipeline.go
+++ b/internal/buildkite/pipeline.go
@@ -158,7 +158,7 @@ func Emit(pipeline Pipeline) ([]byte, error) {
 			if workflow.CheckName == "" {
 				return nil, fmt.Errorf("workflow %q requires a GitHub Check name", workflow.GroupKey)
 			}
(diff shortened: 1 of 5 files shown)


sj26:
Released in [v0.13.3](https://github.com/buildkite/buildkite-gha/releases/tag/v0.13.3)

PR #205, opened 2026-08-14
Override GITHUB_SERVER_URL for actions/cache to avoid GHES misdetection

isaacsu:
## Problem

In the Cursor origin, `actions/cache` logs and skips caching:

```
[warning]Cache action is only supported on GHES version >= 3.5. If you are on version >=3.5 Please check with GHES admin ...
```

## Root cause

`actions/cache`'s bundled `isGhes()` decides "is this GHES?" purely from the `GITHUB_SERVER_URL` hostname, allowlisting only `github.com`, `*.ghe.com`, and `*.localhost`. The Cursor origin sets `GITHUB_SERVER_URL=https://origin.cursor.com`, which matches none of those, so `isGhes()` returns true. That forces the action onto the unsupported cache-v1 path, even though the Buildkite runtime only provisions a cache-v2 backend (`ACTIONS_CACHE_SERVICE_V2`, `ACTIONS_RESULTS_URL`, `ACTIONS_RUNTIME_TOKEN`), so the availability check fails and caching is skipped entirely.

## Fix

Override `GITHUB_SERVER_URL` to a synthetic `.localhost` host (`https://buildkite-gha.localhost`), but only for the cache action's own subprocess environment, and only when the real value would be misdetected as GHES:

- `shouldOverrideGitHubServerURL` mirrors `actions/cache`'s own hostname allowlist to detect misdetection.
- `isolateCacheActionEnvironment` applies the override only when needed, so cache-compatible actions like `setup-go` (which share `cacheActionEnvironment` but not this isolation path) keep their genuine `GITHUB_SERVER_URL`.

This is scoped narrowly: `GITHUB_SERVER_URL` is referenced only by `isGhes()` inside the cache action's dist bundles, and cache traffic targets `ACTIONS_RESULTS_URL`, so the non-resolvable `.localhost` host is never contacted.

## Tests

- `TestShouldOverrideGitHubServerURL`: table test covering github.com, `*.ghe.com`, `*.localhost`, the Cursor origin, real GHES, empty, and malformed inputs.
- `TestIsolateCacheActionEnvironmentOverridesGitHubServerURLWhenNeeded`: confirms the override applies and existing isolation guarantees (PATH forced, secrets stripped) still hold.
- `TestIsolateCacheActionEnvironmentLeavesRealGitHubServerURLUnchanged`: confirms `github.com` passes through untouched.

## Verify

```
mise exec -- go test ./internal/runtime/...
mise run check
```

Both green locally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 dist/
 .release/
+.idea/
 
 # This repository intentionally tracks its reviewed DeepSec configuration.
 !/.deepsec/
--- a/internal/runtime/cache_service.go
+++ b/internal/runtime/cache_service.go
@@ -18,6 +18,12 @@ const (
 	cacheCredentialResponseLimit = 64 << 10
 	cacheActionToolPath          = "/usr/local/bin:/usr/bin:/bin"
 	defaultCacheResultsURL       = "https://ghacs.buildkite.com/"
+
+	// githubServerURLOverride is the synthetic, non-resolvable server URL
+	// substituted in when shouldOverrideGitHubServerURL reports true, so
+	// actions/cache's isGhes() check doesn't force the unsupported cache-v1
+	// path. Only its ".localhost" suffix is load-bearing.
+	githubServerURLOverride = "https://buildkite-gha.localhost"
 )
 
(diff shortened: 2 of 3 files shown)


sj26:
Released in [v0.13.4](https://github.com/buildkite/buildkite-gha/releases/tag/v0.13.4)

PR #208, opened 2026-08-14
Fix cache detection in setup actions

isaacsu:
## Summary

Workflows from non-`github.com` repositories can expose a `GITHUB_SERVER_URL` that the cache clients bundled in setup actions misclassify as GHES. They then fail to use the Buildkite cache-v2 service despite receiving its endpoint and job-bound token.

This extends the workaround from #205 so the built-in dependency caches in root setup-node, setup-java, setup-python, setup-go, and setup-dotnet actions work for those repositories. It:

- shares one descriptor-driven, subprocess-scoped hostname override with direct `actions/cache`
- preserves the normal setup-action environment and real workflow expression context
- applies across versions while excluding nested and unlisted actions
- documents the source and request-path audit required before adding an action

## Verification

- `mise run check`


sj26:
Released in [v0.13.5](https://github.com/buildkite/buildkite-gha/releases/tag/v0.13.5)

PR #214, opened 2026-08-14
PB-2910: apply hosted-toolchains image to unmapped Linux runner labels

zhming0:
Context: [PB-2910](https://linear.app/buildkite/issue/PB-2910)

Unconfigured `ubuntu-latest` / `ubuntu-24.04` / `ubuntu-22.04` labels previously compiled to steps with neither `agents.queue` nor `image`. Jobs landing on the default Hosted Linux queue ran without the agent-base toolchains (`yarn: command not found`), and the only fix was an explicit `runners` mapping.

## Change

Hosted upload defaults now carry the matching hosted-toolchains image while keeping default agent targeting:

| Label | Queue | Image |
| --- | --- | --- |
| `ubuntu-latest`, `ubuntu-24.04` | default targeting | Noble agent-base |
| `ubuntu-22.04` | default targeting | Jammy agent-base |

Explicit `runners` mappings override as before. macOS behavior is unchanged. The standalone `compile` command is unchanged; only the hosted upload/plugin path is affected.

Steps that get the image also get `--hosted-tool-cache`, matching configured profiles.

## Behavior notes

- This assumes GHA pipelines run on Hosted Linux agents (team consensus; self-hosted is out of scope for this product).
- `runs-on: [ubuntu-latest, ubuntu-22.04]` is now rejected as conflicting targets (Noble vs Jammy). Previously both resolved to the same empty target. That label combination never matches a real GitHub-hosted runner either.

## Testing

- `mise run check` passes (format, build, test, race, lint, vet, smoke:local, release:check).
- New `TestRunUploadDefaultsHostedToolchainImageWithoutRunnerQueue` covers the no-config upload path.
- Updated `TestRunnerProfileDoesNotAffectOtherLinuxLabels` for the new default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,9 +70,9 @@ plugins:
           queue: macos-sonoma-arm64
 ```
 
-Configured Linux profiles use the matching Noble or Jammy hosted-toolchains
-image. A macOS label selects native Darwin/arm64, not a GitHub image or Xcode
-inventory.
+Linux labels use the matching Noble or Jammy hosted-toolchains image, with or
+without a configured queue. A macOS label selects native Darwin/arm64, not a
+GitHub image or Xcode inventory.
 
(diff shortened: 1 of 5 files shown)


wolfeidau (approved):
YES 👍🏻 🚀

sj26:
Released in [v0.13.6](https://github.com/buildkite/buildkite-gha/releases/tag/v0.13.6)

PR #225, opened 2026-08-14
PB-2722: Default macos-latest to the hosted macos-medium queue

zhming0:
Part of [PB-2722](https://linear.app/buildkite/issue/PB-2722). Closes the `macos-latest` auto-routing gap discussed in [#project-buildkite-gha](https://buildkite-corp.slack.com/archives/C0BKA9NUYS0/p1786683008012169).

## What

Unmapped `runs-on: macos-latest` in the hosted upload path now targets the `macos-medium` queue that Buildkite provisions at signup, instead of failing compilation with an unmapped-label diagnostic.

| Label | Without a mapping | With a mapping |
| --- | --- | --- |
| `macos-latest` | queue `macos-medium` | configured queue |
| `macos-14`, `macos-15` | rejected (unchanged) | configured queue |

- Versioned macOS labels keep requiring an explicit queue: no default queue can guarantee a specific macOS or Xcode version. `macos-latest` floats on GHA too, so routing it to a floating queue is honest.
- The default queue joins the untrusted-queue allowlist the same way configured queues always have.
- Scope is `hostedOptions` only (`upload` and `validate --profile hosted`). Standalone `compile` and plain `validate` are unchanged, matching how PB-2910 scoped the Ubuntu image defaults.
- No runtime-distribution changes: the plugin acquisition path already fetches the Darwin runtime when a workflow resolves to darwin/arm64.

## Caveat

The failure mode for `macos-latest` shifts from a compile-time diagnostic to a job waiting on a queue with no agents when an org renamed `macos-medium` or has no macOS capacity (for example, after eval expiry). Documented in `docs/cli.md`; the queue-discovery API is the longer-term fix.

## Testing

- New `TestCompileHostedDefaultsMacOSLatestToHostedQueue`: unmapped `macos-latest` still requires the Darwin runtime distribution, compiles to `queue: "macos-medium"` when present, and an explicit mapping overrides the default.
- `mise run check` passes (format, build, test, race, lint, vet, local smoke, release check).

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,8 +71,9 @@ plugins:
 ```
 
 Linux labels use the matching Noble or Jammy hosted-toolchains image, with or
-without a configured queue. A macOS label selects native Darwin/arm64, not a
-GitHub image or Xcode inventory.
+without a configured queue. Without a mapping, `macos-latest` targets the
+hosted `macos-medium` queue; `macos-14` and `macos-15` require one. A macOS
+label selects native Darwin/arm64, not a GitHub image or Xcode inventory.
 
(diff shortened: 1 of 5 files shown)


sj26:
Released in [v0.13.10](https://github.com/buildkite/buildkite-gha/releases/tag/v0.13.10)

PR #226, opened 2026-08-14
Lint checkout snapshot generator

lox:
## Why

The checkout main-history snapshot generator added in #215 is not included in the repository's shell lint task, so regressions in it can bypass `mise run check`.

## What

Add `scripts/update-checkout-main-commits` to the existing ShellCheck invocation.


sj26:
Released in [v0.13.12](https://github.com/buildkite/buildkite-gha/releases/tag/v0.13.12)

PR #229, opened 2026-08-14
Link workflow token errors to pipeline settings

lox:
## Why

A disabled workflow-token request currently says only that the organization or pipeline is not enabled. The organization rollout can already be active while the required per-pipeline repository setting remains off, leaving users without the actual fix.

## What

Tell users to enable **Allow workflow-authorized GitHub access tokens** and link directly to the affected pipeline's repository settings. The URL is included only when the immutable Buildkite organization and pipeline slugs are safe canonical values; otherwise the instruction remains without a link.

For example:

```text
buildkite-gha: run-job: GitHub workflow access tokens are not enabled for this organization or pipeline; enable "Allow workflow-authorized GitHub access tokens" in the pipeline's repository settings: https://buildkite.com/acme-inc/my-pipeline/settings/repository
```


sj26:
Released in [v0.13.13](https://github.com/buildkite/buildkite-gha/releases/tag/v0.13.13)

PR #230, opened 2026-08-14
Add Test Engine analytics for Go tests

lox:
## Why

The Go suite has no per-test analytics, and two packages dominate its runtime. A cancellation test also waits for a one-second deadline instead of synchronizing on the behavior it tests.

## What

- Run Linux Go tests through Buildkite Test Engine with two package-level partitions and JSONL result upload.
- Keep race tests in the repository checks without repeating the standard suite there.
- Make artifact cancellation deterministic and run isolated artifact boundary tests in parallel, reducing an uncached local suite from 28.9s to 15.3s.


sj26:
Released in [v0.13.14](https://github.com/buildkite/buildkite-gha/releases/tag/v0.13.14)

PR #231, opened 2026-08-14
Make plugin import output concise and failure-first

lox:
## Why

Plugin imports currently print the complete processing report, so passed stage, job, and action inventories bury the result and make admission failures harder to find. Public `validate` and direct `upload` output are stable contracts and should remain verbose.

## What

Use a plugin-only presentation boundary. Successful imports now keep the existing one-line result:

```text
Uploaded 3 jobs from 1 workflows using /path/to/buildkite-gha with importer plugin-linux.
```

Admission failures open the Buildkite log group and lead with the actionable diagnostic and its workflow, stage, and job context:

```text
^^^ +++
x [E_PROFILE] Job "secret" uses GitHub Actions secrets, which hosted runs do not provide. Pass values through supported Buildkite secret handling or remove the dependency. (./.github/workflows/secret.yml:4:3) {workflow=.github/workflows/secret.yml, stage=hosted-profile-admission, job=secret}
Compilation: compilable. Admission: not-admitted.
```

The existing job-scoped annotation continues to carry full diagnostic detail and source links. Public `validate --format text|json` and direct `upload` still use the stable verbose renderer.

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -1498,6 +1498,7 @@ func uploadParsedContext(ctx context.Context, uploadArguments parsedUploadArgs, 
 		}
 	}
 	out := newProcessingOutput("upload", "text", stderr, stderr, agent)
+	out.plugin = uploadArguments.pluginAcquisition != nil
 	if uploadArguments.telemetry != nil {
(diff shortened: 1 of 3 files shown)


sj26:

PR #235, opened 2026-08-14
Use latest GitHub Actions plugin and mise cache in examples

lox:
## Why

The example GitHub Actions importer caches mise-managed runtimes, but its cache volume still uses the old plugin-specific path and its plugin selector is pinned to an older release.

## What

Use `/cache/bkcache/mise` and `github-actions#latest` for the importer in both the checked-in example pipeline and the dynamically generated pipeline.

Related: https://github.com/buildkite/buildkite/pull/32603 adds the same cache to steps generated during new-pipeline workflow detection.

### DIFF
--- a/.buildkite/examples.yml
+++ b/.buildkite/examples.yml
@@ -67,7 +67,7 @@ steps:
               timeout_in_minutes: 20
               retry:
                 automatic: false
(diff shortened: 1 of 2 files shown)


wolfeidau (approved):
👍🏻

sj26:

PR #238, opened 2026-08-15
Support contains and endsWith in compile-time expressions

lox:
## Why

The narrow compile-time expression surfaces support `startsWith()` but reject the equivalent string checks workflows commonly express with `contains()` and `endsWith()`.

## What

Add case-insensitive string `contains()` and `endsWith()` evaluation alongside `startsWith()` when expressions resolve fully during compilation. Keep wrong arities, non-string arguments, runtime conditions, and general runtime interpolation fail-closed; GitHub's array form of `contains()` remains unsupported.

Update the compatibility reference and add focused coverage for compile-time conditions, case-insensitive matches, negative results, and rejected argument shapes.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -225,7 +225,7 @@ defaults:
 
 ### Concurrency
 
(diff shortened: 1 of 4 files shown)


sj26:

PR #243, opened 2026-08-15
Deduplicate native checkout policy

lox:
## Why

The native checkout path accumulated parallel version checks, duplicate SHA validation, and repeated ref classification while adding bounded v3 support.

## What

Centralize the v3 contract predicate and exact-SHA policy in the checkout integration, then classify the selected checkout revision once for shallow fetches, full-history fetches, and detached checkout targets. Preserve the existing security-focused test setups and behavior.


sj26:

PR #244, opened 2026-08-15
Centralize action integration admission

lox:
## Why

Action identity classification and audited-commit admission were separate operations, so compiler, runtime, and CLI callers had to pair each descriptor with the correct validator. The package also kept checkout, artifact, cache, diagnostics, and shared input policy in one file.

## What

Organize the integration policy into catalog, admission, and action-focused files. Add atomic identity-plus-commit admission, route resolved action locks through it, and keep the action-specific commit validators private while preserving existing runtime execution and compatibility behavior.


catkins (approved):

sj26:

PR #246, opened 2026-08-15
Fix legacy cache v2 availability detection

lox:
## Why

The audited `actions/cache` v3 releases selected cache v2 but could not reach Buildkite Results. Their bundled client still checks `ACTIONS_CACHE_URL` as an availability gate before using `ACTIONS_RESULTS_URL`, so the action exited before restore or save.

## What

Provide the existing non-routable cache availability sentinel to audited cache actions while keeping all traffic and job-scoped credentials on `ACTIONS_RESULTS_URL`. Record the hosted proof from Buildkite build 1173, where v3.4.0 saved a build-unique archive and v6.1.0 restored and verified it.


sj26:

PR #248, opened 2026-08-15
Reduce Test Engine parallelism

lox:
## Why

Test Engine splits Go tests by package. `internal/runtime` takes about 24 seconds by itself, while the other 13 packages finish together in about 5 seconds, so a second agent does not shorten the critical path.

## What

Run the Test Engine step with one partition. This preserves Test Engine collection and duration history while avoiding an idle second agent.


sj26:

PR #250, opened 2026-08-15
Split action integration tests by concern

lox:
## Why

The action integration implementation is organized by admission, catalog, and individual actions, but its tests remain in one monolithic file. This makes each action's policy and coverage harder to inspect together.

## What

Split the existing tests into files matching the implementation concerns. Keep every assertion and test case unchanged.


sj26:

PR #251, opened 2026-08-15
Validate workflows with generated events

lox:
## Why

Users currently need to craft a complete event snapshot before checking hosted workflow compatibility, while bare validation can miss trigger translation failures that upload rejects.

## What

Validate every declared trigger during event-independent validation, and add generated snapshots through `--event` for one event or `--all-events` for every declared supported event. Single-event reports remain v2; all-event validation uses a v3 aggregate that preserves the event-independent report and each event-specific admission result without treating generated payloads as real webhook evidence.


lox:
Acknowledged reviews 4944854698 and 4944885722: no defects to address. The intended v3 per-event semantics were reviewed in the parent thread, the repository check suite passed, and the owner authorized merge when checks are green.

sj26:

PR #254, opened 2026-08-16
Cache mutable action ref resolutions

lox:
## Why

Corpus validation repeatedly resolves the same mutable action refs through GitHub, consuming API quota and time even when another trusted worker resolved them recently.

## What

Cache mutable ref resolutions for one hour in the platform XDG cache. Per-ref file locks coalesce concurrent validation workers, and atomic bounded entries preserve a clear revalidation window when tags or branches move.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -70,7 +70,7 @@ buildkite-gha validate \
   .github/workflows/ci.yml
 ```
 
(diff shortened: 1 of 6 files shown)


lox:
@buildsworth-bk review

The macOS-only cache path assertion now derives its expectation from `os.UserCacheDir`; the prior test incorrectly assumed macOS honors `XDG_CACHE_HOME`.

sj26:

PR #255, opened 2026-08-16
Batch public corpus validation

lox:
## Why

The public workflow corpus is too large to launch one validator process per workflow, and interrupted runs currently restart completed work.

## What

Add a concurrent batch validation command that shares action resolution across workflows and atomically resumes results keyed by corpus record, content, and validator executable. Update the corpus script to use the batch path while retaining one `processing-report/v3` result per workflow.


lox:
Addressed for review 4945300582: indeterminate reports are retried rather than resumed.

Addressed for review 4945305972: the resume semantics were fixed in the subsequent code change.

Addressed for review 4945309936: the follow-up found no new concerns.

Addressed for review 4945366752: the rebased revision has no new concerns and all checks pass.

sj26:

PR #256, opened 2026-08-16
Persist indeterminate batch reports

lox:
## Why

`validate-batch` aborts a corpus run when hosted validation produces a structurally valid `indeterminate` report, which is expected for workflows whose referenced local actions are absent from the corpus snapshot.

## What

Persist structurally valid indeterminate reports so the current batch can finish, while continuing to reject them as resumable results so a later run retries them.


sj26:

PR #258, opened 2026-08-16
Fix trigger validation call

lox:
## Why

The merged path-filter change added a required argument to the trigger translator but left one validation call unchanged, so `main` does not compile.

## What

Mark each trigger as selected when validating it against the same live-context rules used by pipeline translation.


sj26:

PR #261, opened 2026-08-16
Add deterministic public corpus sampling

lox:
## Why

Running the full public workflow corpus is too slow for compatibility hill climbing. Developers need a stable subset that remains comparable across validator commits without mixing its reports with full-corpus results.

## What

Add deterministic `SAMPLE_SIZE` and `SAMPLE_SEED` selection based on repository, workflow path, and content hash. Give each selection an isolated manifest, corpus ID, report directory, and tally keyed by its sample identity.

Document the corpus benchmark, recommended sample tiers, caching and snapshot behavior, outputs, refresh semantics, and reproducibility limits in the development guide.


sj26:

PR #269, opened 2026-08-16
Make background task lifecycle explicit

lox:
## Why

Background tasks track lifecycle with three booleans whose combinations can represent invalid states and obscure cancellation and commit invariants.

## What

Replace the booleans with one private, mutex-protected lifecycle state that allows only queued, running, finished, and committed transitions. Keep completion synchronization, context cancellation, dispatch ordering, and the supervisor call surface unchanged.

Add focused coverage for valid and impossible transitions and exactly-once commit under concurrent barriers.

### DIFF
--- a/internal/runtime/concurrent.go
+++ b/internal/runtime/concurrent.go
@@ -23,6 +23,15 @@ type stepExecution struct {
 	conclusion string
 }
 
+type backgroundTaskState uint8
+
+const (
+	backgroundTaskQueued backgroundTaskState = iota
+	backgroundTaskRunning
+	backgroundTaskFinished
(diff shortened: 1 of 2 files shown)


lox:
Acknowledged for review 4947422154: no actionable findings require a branch change. The state-transition and cancellation/commit interleavings were traced in the local review and Oracle pass, and the focused race coverage and CI are green. Final human approval remains outside this babysitting pass.

wolfeidau (approved):
👍🏻

sj26:
Released in [v0.15.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.15.0)

PR #277, opened 2026-08-17
Use one completed-step runtime record

lox:
## Why

Runtime expression evaluation stores completed-step outputs and outcome/conclusion in parallel maps. Every completion path must update both representations, which obscures ownership and allows the values to diverge.

## What

Use `expression.StepStatus` records in `expression.Context.Steps` as the single source for direct references, whole `steps` objects, conditions, controls, background barriers, and composite children. Keep `Context` and `ConditionContext` phase-specific while preserving case-insensitive lookup, missing-value behavior, continue-on-error semantics, barrier visibility, and composite isolation.

Deep-copy step output maps in context clones, with focused coverage for skipped and control records and composite child state. An Oracle review traced the ordinary, skipped, control, background, composite, and continue-on-error paths and found no behavior change.

### DIFF
--- a/internal/expression/expression.go
+++ b/internal/expression/expression.go
@@ -33,8 +33,7 @@ type Expression struct {
 	Span Span   `json:"span"`
 }
 
-// StepStatus contains the values exposed for one completed step while
-// evaluating a condition.
+// StepStatus contains the runtime values exposed for one completed step.
 type StepStatus struct {
 	Outcome    string
 	Conclusion string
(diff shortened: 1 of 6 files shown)


lox:
Acknowledged for review 4947966064: no actionable findings. The exact head has passed the local risk-matched review, Oracle flow trace, full repository check, and branch CI; no branch change is required.

sj26:

PR #278, opened 2026-08-17
Use one prerequisite expression record

lox:
## Why

Expression evaluation stores prerequisite outputs and results in parallel maps. Construction, cloning, and resolution must keep both representations aligned, obscuring that each result and its outputs belong to one prerequisite.

## What

Add an expression-owned `NeedStatus` record and use it in the distinct runtime and condition contexts. Project each verified `plan.Need` once, copying only its result and outputs; verified artifact authority remains runtime-only on `plan.Need` and never enters expression state.

Update whole and direct `needs` access, job and step conditions, and context cloning while preserving case-insensitive lookup, missing-value behavior, prerequisite conclusions and validation ordering, reusable and matrix output projection, and the existing expression surface. Focused tests cover lookup semantics and map isolation, the full check passes, and an Oracle review traced these boundaries without finding a behavior change.

### DIFF
--- a/internal/expression/condition.go
+++ b/internal/expression/condition.go
@@ -13,8 +13,7 @@ import (
 // job or step condition.
 type ConditionContext struct {
 	Inputs       map[string]any
-	Needs        map[string]map[string]string
-	NeedResults  map[string]string
+	Needs        map[string]NeedStatus
 	Steps        map[string]StepStatus
 	Env          map[string]string
 	Vars         map[string]string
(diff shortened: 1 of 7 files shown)


lox:
Acknowledged for review 4948114338: Buildsworth found no correctness issue and raised no actionable finding. The exact head also passed the staged Oracle trace, local risk-matched ship/simplicity review, full repository checks, and PR CI. No branch change is required; explicit human merge approval remains pending.

sj26:

PR #279, opened 2026-08-17
Apply hosted runners to all-events validation

lox:
## Why

`validate --profile hosted --all-events` performed its event-independent pass with the default Linux-only runner policy. It therefore rejected `macos-latest` before reaching hosted evaluation, even though the hosted preset supports that label.

The deterministic 10K corpus contained 600 workflows with this false diagnostic; it was the sole reported error for 164.

## What

Add event-independent validation with explicit compiler options and use the hosted runner policy for the aggregate all-events pass. Bare validation retains its existing default policy.

Add CLI coverage proving a case-insensitive `macOS-latest` workflow completes both aggregate validation and generated hosted event admission. In an exact 10K comparison, the false diagnostic fell from 600 workflows to zero, 73 workflows became admitted, and no workflow regressed.


sj26:

PR #282, opened 2026-08-17
Explain security boundaries for GitHub Actions users

lox:
## Why

The security model starts in Buildkite terms, so GitHub Actions users must translate familiar runner, credential, and identity boundaries before applying its guidance.

## What

Introduce the model in GitHub Actions terms and add a concise mapping for runners, jobs, `GITHUB_TOKEN`, secrets, and OIDC. Keep the detailed operator guidance unchanged.


wolfeidau (approved):
much nicer 👍🏻

sj26:
Released in [v0.30.1](https://github.com/buildkite/buildkite-gha/releases/tag/v0.30.1)

PR #285, opened 2026-08-17
Disable the Windows starter workflow corpus case

lox:
## Why

The pinned CMake starter workflow includes a `windows-latest` matrix row. The hosted profile does not support Windows, so this known platform gap blocks the otherwise compatible starter corpus.

## What

Keep the CMake template pinned but disable it by default with a recorded reason. Report disabled cases in the corpus summary and annotation, while allowing explicit case selection to recheck them when Windows support changes.


sj26:

PR #289, opened 2026-08-17
Clarify action admission diagnostics and expand runtime proof

lox:
## Why

Hosted admission validates action source and metadata without executing action code, but the repeated third-party runtime warning makes successful starter-workflow validation look like an error. Normal CI also admitted well-known starter workflows without executing the existing public-action runtime proof.

## What

Describe the limitation as unevaluated action runtime behavior and retain it in processing reports without publishing it as a Buildkite warning annotation. Replace repeated starter-workflow warnings with one admission caveat.

Extend the pinned public-actions proof with Python 3.13 through setup-python v7 and Temurin Java 21 through setup-java v5, then run that proof in normal CI against the build's exact CLI source.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -123,6 +123,11 @@ steps:
               version: "2026.5.12"
         command: "mise run corpus:starter-profile"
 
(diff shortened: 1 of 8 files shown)


sj26:

PR #290, opened 2026-08-17
Consolidate action source lock mechanics

lox:
## Why

Action-cache locks and mutable-reference and snapshot locks duplicate lock-file opening, flock polling, cancellation, and cleanup even though their policies and platform coverage differ. Keeping those mechanics separate makes behavior-preserving maintenance harder and creates room for the implementations to drift.

## What

Extract private lock-file ownership and Unix flock polling while retaining policy-specific wrappers:

| Platform | Action cache | Mutable references and snapshots |
| --- | --- | --- |
| Linux, Darwin | Shared/exclusive flock | Exclusive flock |
| FreeBSD | Shared/exclusive flock | Open file only; no exclusion |
| Windows | Process-local shared/exclusive `sync.RWMutex` | Open file only; no exclusion |

Preserve `0600` lock files, 10 ms action-cache and 25 ms mutable-reference polling, cancellation and error identities, nonblocking eviction semantics, lease lifetime, and close-once idempotent cleanup. Focused source and race tests, four-platform test-binary cross-compilation, and the full repository check pass; an Oracle review traced the build tags, policy paths, lock lifetime, polling, and cleanup without finding a behavior regression.

### DIFF
--- a/internal/action/source/action_cache_lock.go
+++ b/internal/action/source/action_cache_lock.go
@@ -1,0 +1,33 @@
+package source
+
+import (
+	"errors"
+	"sync"
+)
+
+type actionCacheLockMode int
+
+const (
+	actionCacheLockShared actionCacheLockMode = iota
+	actionCacheLockExclusive
+)
+
+var errActionCacheLockUnavailable = errors.New("action cache lock is held")
+
+type actionCacheLock struct {
+	file    *lockFile
+	release func()
+	once    sync.Once
+}
+
+func (l *actionCacheLock) unlock() {
+	if l == nil {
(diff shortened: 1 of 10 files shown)


lox:
Acknowledged for review 4949980098: no inline findings or branch changes were requested. A risk-matched local review and the pre-publication Oracle trace both checked the concurrency, build-tag, lifetime, polling, cancellation, and cleanup paths without finding a behavior regression. All local validation, cross-compilation, and PR checks pass; the PR is being held for explicit user merge approval.

sj26:

PR #294, opened 2026-08-17
Admit event-derived GitHub ref scalars at compile time

lox:
## Why

The compiler already derives `github.ref_name`, `github.ref_type`, and `github.base_ref`, but its compile-time whitelist rejects them in matrices, runner labels, names, concurrency groups, and event-backed condition reduction.

## What

Admit those three event-derived scalars through the existing compile-time expression surface and fold condition references before plan emission. Keep the existing ref and PR base-branch derivation and runner-policy boundary unchanged, with focused push and pull-request coverage. Document the complete compile-time GitHub context.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -592,7 +592,7 @@ Patterns cannot be absolute, contain a `..` path segment, or contain ASCII contr
 
 ### Compile-time expressions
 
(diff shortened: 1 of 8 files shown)


lox:
Acknowledged for review 4950440979: no actionable findings. The condition-folding path was manually checked against push, tag, pull-request, pull-request-target, and missing-base-ref cases; the emitted-plan assertions verify these compile-time scalars do not cross the runtime boundary.

lox:
Superseded by #295, which contains the corrected implementation and has merged.

PR #295, opened 2026-08-17
Admit event-derived GitHub ref scalars at compile time

lox:
## Why

The compiler already derives `github.ref_name`, `github.ref_type`, and `github.base_ref`, but its compile-time whitelist rejects them in matrices, runner labels, names, concurrency groups, and event-backed condition reduction.

The first implementation in #294 also evaluated expression-looking static matrix values, which caused unrelated `E_MATRIX_INVALID` regressions. The corrected implementation's deterministic 10K A/B moved 18 workflows out of incompatible with no regressions and left `E_MATRIX_INVALID` unchanged.

## What

Admit only those three event-derived scalars through the existing compile-time expression surface and fold their condition references before plan emission. Preserve existing ref derivation, static matrix expansion, unsupported runtime fields, and runner-policy enforcement.

Add focused push, pull-request, matrix, policy, and regression coverage, and document the complete compile-time GitHub scalar surface.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -594,7 +594,7 @@ Patterns cannot be absolute, contain a `..` path segment, or contain ASCII contr
 
 ### Compile-time expressions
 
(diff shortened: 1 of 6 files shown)


sj26:

PR #296, opened 2026-08-17
Document initial compatibility scope boundaries

lox:
## Why

Raw compatibility results currently mix product-scope decisions with gaps we intend to close. Windows execution requires a real Windows target and is not part of the initial Linux and Apple-silicon macOS product scope.

## What

Document Windows execution as outside the initial scope and retain fail-closed runner-label behavior. Clarify that corpus reports must keep Windows in raw totals and compute any additional in-scope view from individual records rather than subtracting overlapping aggregate counts.

This changes documentation and benchmark interpretation only; validation and runtime behavior are unchanged.


sj26:

PR #299, opened 2026-08-17
Support conditions on local reusable workflow calls

lox:
## Why

Local reusable-workflow calls currently reject job-level `if`, even though GitHub evaluates that condition in caller scope before expanding the call. This blocks otherwise compatible workflows and cannot be implemented safely by splicing the caller expression into callee job conditions.

## What

Carry a bounded, ordered call-guard stack through reusable-workflow expansion and immutable job plans. Reduce and validate each guard against caller `github`, `vars`, `inputs`, direct `needs`, and status functions, with independently verified producer manifests and Buildkite dependencies.

Evaluate guards before the callee job condition and before credentials, tokens, OIDC, actions, containers, or steps. A false outer guard publishes skipped descendants with empty outputs, including descendants with `always()`, while nested guards retain outer-to-inner ordering and callee failures cannot alter caller guard state.

Add compiler, plan/schema, runtime, transport, and CLI coverage, and document the compatibility and security boundaries. The deterministic corpus has not been rerun, so this PR makes no realized compatibility-impact claim.


lox:
Acknowledged for review 4955106567: no actionable findings; no changes needed.

sj26:
Released in [v0.19.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.19.0)

PR #301, opened 2026-08-18
PB-2962: Support compound expressions in composite run steps

zhming0:
## Root cause

[starter-workflows-unified-test build 10](https://buildkite.com/buildkite-github-actions/starter-workflows-unified-test/builds/10) reaches `actions-rust-lang/setup-rust-toolchain@v1` after the PB-2962 lifecycle-expression work, then fails in composite step 2 (`id: flags`) on:

```yaml
${{ contains(inputs.toolchain, 'nightly') && inputs.components && ' --allow-downgrade' || '' }}
```

Composite action `run` fields used the direct-reference-only runtime evaluator. The function call and logical operators therefore returned `unsupported expression reference`, even though GitHub Actions supports this expression surface and ordinary workflow `run` fields already use the full step evaluator.

## Implementation

Evaluate composite action `run` scripts with the existing step-expression evaluator. This supports compound expressions without broadening unrelated expression surfaces. Composite metadata still cannot call `hashFiles()`; its error is now the specific unavailable-function diagnostic.

## Regression coverage

- Added an end-to-end composite action runtime test using the exact `setup-rust-toolchain` expression.
- Verifies the nightly toolchain and components inputs resolve to ` --allow-downgrade`.
- Updated the existing composite `hashFiles()` restriction assertion for the more specific diagnostic.

## Verification

`mise run check` passed in full:

- formatting
- Linux and macOS builds
- unit tests
- race tests
- Go lint and vet
- shellcheck
- local smoke tests
- release configuration checks

### DIFF
--- a/internal/runtime/hash_files_test.go
+++ b/internal/runtime/hash_files_test.go
@@ -237,7 +237,7 @@ func TestHashFilesRemainsUnavailableOutsideWorkflowStepFields(t *testing.T) {
 
 	writeFixtureFile(t, workspace, ".github/actions/composite/action.yml", "runs:\n  using: composite\n  steps:\n    - shell: sh\n      run: echo \"${{ hashFiles('value') }}\"\n")
 	job = runtimePlan(t, workspace, ".github/workflows/test.yml", []plan.Step{{ID: "composite", Kind: "uses", Uses: "./.github/actions/composite"}})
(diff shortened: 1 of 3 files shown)


sj26:
Released in [v0.19.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.19.0)

PR #308, opened 2026-08-18
Remove unreachable Node 20 execution state

lox:
## Why

Action metadata declares `node20`, but classification already normalizes it to the managed Node 24 runtime. Retaining a separate Node 20 execution value leaves impossible branches in entrypoint validation, compilation, and lifecycle dispatch.

## What

Keep `node20` as a private accepted declaration spelling that maps to the typed Node 24 execution runtime. Remove the unreachable Node 20 runtime value and its downstream branches without changing supported declarations, Node selection, or lifecycle behavior.


sj26:

PR #314, opened 2026-08-18
PB-2989: Support github.action_path and expand the runtime github context subset

zhming0:
## Why

A composite action step that references `${{ github.action_path }}` fails at runtime:

```
buildkite-gha: run-job: step "step-2": composite action step 4: unsupported runtime github reference "github.action_path"
```

The value already exists — composite steps export `GITHUB_ACTION_PATH` — but the step-expression whitelist rejects the expression form. The whitelist was also stricter than the runtime subset documented in `docs/compatibility.md`: `github.base_ref`, `github.ref_name`, `github.ref_type`, and `github.repository_owner` are documented as retained but were rejected in step fields.

Linear: PB-2989 (related to the PB-2725 umbrella card).

## What

The runtime populates `github.action_path` per composite invocation with the same scoping as `GITHUB_ACTION_PATH`; nested composites overlay their own path on entry. Outside composite actions the expression resolves to an empty string, matching GitHub.

The step-runtime `github` whitelist now admits the documented members plus `github.workflow` and `github.job`, whose values are already exported as `GITHUB_WORKFLOW` and `GITHUB_JOB`. `run_id`, `run_number`, `run_attempt`, `triggering_actor`, `action_ref`, `action_repository`, and `github.event` remain unsupported and fail closed.

An integration test runs a nested composite fixture asserting expression and environment-variable agreement at each nesting level, and emptiness at the workflow level.


sj26:
Released in [v0.21.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.21.0)

PR #316, opened 2026-08-18
Evaluate composite step working-directory expressions

zhming0:
## Why

Running a plain starter Go workflow failed with:

```
buildkite-gha: run-job: step "step-2": composite action step 7: fork/exec /usr/bin/bash: no such file or directory
```

The workflow was fine. A composite action step (`actions-rust-lang/setup-rust-toolchain@v1` step 7 in the reproduction; `actions/setup-go` pulls in a similar pattern) declared `working-directory: ${{ inputs.rust-src-dir || '.' }}`. The runtime used the expression text literally as a path, the child process failed to chdir, and Go surfaced that as a misleading `fork/exec <shell>` error.

## What

- Evaluate composite action step `working-directory` with the step expression evaluator, matching the existing handling of `run`, `env`, and `with`.
- Resolve run-step working directories through a shared `stepWorkingDirectory` helper that requires the directory to exist, so a bad path now fails with `working directory "<path>" does not exist in the workspace` instead of a `fork/exec` error.
- Document composite `working-directory` expression support in `docs/compatibility.md`.

Covered by `TestCompositeStepWorkingDirectoryEvaluatesExpressions` and `TestRunStepMissingWorkingDirectoryFailsClearly`.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -583,7 +583,7 @@ Job-level expressions support the same operators and pure functions with these f
 | `defaults.run` | `github`, `needs`, `matrix`, `env`, `vars`, `inputs` |
 | `outputs` | `github`, `needs`, `matrix`, `runner`, `env`, `vars`, `secrets`, `steps`, `inputs` |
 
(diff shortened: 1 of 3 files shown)


sj26:
Released in [v0.23.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.23.0)

PR #323, opened 2026-08-19
Enable context propagation lint

wolfeidau:
## Why

Nested operations created background contexts that discarded caller values and made cancellation ancestry difficult to verify.

## What

Enable `contextcheck` and propagate caller contexts through CLI, action-source, telemetry, and runtime paths. Cleanup, publication, and telemetry that must outlive cancellation use `context.WithoutCancel` so they retain context values while keeping their existing bounded lifetimes.


zhming0 (approved):

sj26:

PR #326, opened 2026-08-19
Refactor runtime execution around per-job sessions

zhming0:
## Why

`Runner` mixed reusable runtime configuration with mutable state for one job, while `RunJob` encoded preparation, action lifecycle, step execution, cleanup, and result publication in one procedure. That made job ownership and runner reuse semantics implicit and increased the risk of lifecycle changes.

## What

- Keep `Runner.RunJob` as the public facade, but create one private `jobRun` for each invocation.
- Move mutable execution state and resources into that session, including evaluation state, containers, action preparations, post actions, background work, artifacts, warnings, and token service lifecycle.
- Decompose execution into preparation, pre-action, step, post-action, and finalization phases while preserving deferred cleanup ordering.
- Separate job results and action lifecycle/execution concerns within the existing `runtime` package, without new interfaces or subpackages.

## Verification

- `mise run check`
- `go test -race ./internal/runtime`

The main risk is lifecycle cleanup ordering. The preparation phase deliberately keeps its deferred resources alive through post actions and finalization, and the full and targeted race suites pass.

### DIFF
--- a/internal/runtime/action_execution.go
+++ b/internal/runtime/action_execution.go
@@ -26,209 +26,6 @@ import (
 	"github.com/buildkite/buildkite-gha/internal/transport"
 )
 
-// JobResult is the bounded logical result returned to the transport layer.
-type JobResult struct {
-	Conclusion         string                     `json:"conclusion"`
-	Outputs            map[string]string          `json:"outputs,omitempty"`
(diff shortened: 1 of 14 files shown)


sj26:

PR #327, opened 2026-08-19
PB-3002: Ignore unsupported trigger events beside supported ones

zhming0:
## Why

The largest single compatibility blocker in the public workflow corpus: a workflow mixing supported and unsupported trigger events, such as `on: [push, issues]`, failed to compile because every declared trigger was validated regardless of which event fired. One `issues` or `pull_request_target` trigger broke CI for the `push` events we fully support.

Decision recorded in [PB-3002](https://linear.app/buildkite/issue/PB-3002): events like `issues`, `issue_comment`, `pull_request_target`, and `workflow_run` stay unsupported for now, but they must not break the build.

## What

An unsupported trigger event has no Buildkite build source and can never start a build, so translation now ignores it instead of failing:

- `TranslateEventTriggerCondition` skips unsupported non-selected triggers. A mixed workflow compiles for its supported events; a workflow declaring only unsupported events becomes not-applicable (a skipped step on upload) instead of a failing step.
- Event-independent validation (`ValidateTriggerConditions`) tolerates unsupported events when the workflow declares at least one supported trigger event, and still rejects a workflow with none.
- The compiler emits a source-located `W_TRIGGER_EVENT_UNSUPPORTED` warning for each ignored trigger, so users still learn that their issue automation will not run. Triggers now carry a parse position to locate the warning.
- `docs/compatibility.md` documents the behavior in the trigger event table.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -26,7 +26,7 @@ Validate syntax, the static graph, and every declared trigger without an event:
 buildkite-gha validate .github/workflows/ci.yml
 ```
 
-This event-independent check accepts syntactically valid push and pull-request path filters because linked-webhook admission can evaluate them with a verified local git diff. It rejects malformed path filters, unsupported events, unsupported branch and tag filter combinations, and unsupported pull-request activity types. It does not resolve actions, evaluate event payload expressions, or claim hosted admission.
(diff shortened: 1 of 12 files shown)


sj26:

PR #334, opened 2026-08-19
Support inherited secrets in local reusable workflows

lox:
## Why

Local reusable workflows could not opt into the existing job-scoped Buildkite secret model, even though flattening already tracked one-hop secret authority and plans already scoped static secret names per expanded job.

## What

Support `secrets: inherit` for repository-local reusable-workflow calls. Each expanded callee job requests only its own statically referenced ordinary Buildkite secrets, and every nested edge must repeat inheritance. Public remote inheritance, explicit mappings, required call-secret declarations, non-static access, and metadata-granted authority remain rejected. `GITHUB_TOKEN` stays on its separate scoped workflow-token contract.

Document that runtime resolution uses the destination job authenticated Buildkite Agent session and that Buildkite Secret access policies remain authoritative.


zhming0 (approved):
LGTM ✅

sj26:
Released in [v0.28.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.28.0)

PR #335, opened 2026-08-19
Accept unknown pull request mergeability

wolfeidau:
## Why

GitHub can report `mergeable: null` while it computes pull request mergeability. We rejected that transient webhook state even when the local checkout contained a valid synthetic merge, producing a misleading path-filter failure.

## What

Accept unknown mergeability when the synthetic merge commit, base, and head pass the existing local verification. Continue rejecting pull requests that explicitly report `mergeable: false`.


sj26:
Released in [v0.27.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.27.0)

PR #341, opened 2026-08-20
Fix compatibility gap report parsing

zhming0:
## Why

The compatibility-gap harness merged `validate` stderr into its JSON report. In Buildkite, automatic annotation progress is written to stderr, so valid admitted reports were followed by non-JSON text and classified as blocked.

## What

Keep validator stderr separate from the machine-readable report, while retaining it for blocked-case diagnostics. Add a Buildkite-mode regression test that emits annotation chatter alongside admitted JSON and verifies the aggregate tally remains accurate.

### DIFF
--- a/internal/cli/compatibility_gaps_script_test.go
+++ b/internal/cli/compatibility_gaps_script_test.go
@@ -1,0 +1,58 @@
+package cli
+
+import (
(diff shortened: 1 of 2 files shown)


sj26:

PR #343, opened 2026-08-20
PB-3030: Support top-level permissions write-all

zhming0:
## Why

GitHub Actions workflows can use top-level `permissions: write-all`, but buildkite-gha rejects it even though the equivalent `read-all` normalization is supported. This blocks 301 repositories in the public corpus.

Related to [PB-3030](https://linear.app/buildkite/issue/PB-3030/support-top-level-permissions-write-all).

## What

Accept top-level `write-all` and normalize it during compilation to the workflow-token endpoint's explicit 13-scope write map. Keep job-level aliases, excluded scopes, pull-request and merge-queue restrictions, and the existing token security boundaries unchanged.

The buildkite server must recognize and deploy the source-policy alias before this client change can merge or release; older server code rejects `write-all` when it verifies the immutable workflow.


zhming0:
@buildsworth-bk did it really expand expands GITHUB_TOKEN write authority? I am looking at it like a syntax sugar support. review again.

sj26:

PR #345, opened 2026-08-20
Document immutable native release commits

lox:
## Why

[buildkite/buildkite#32685](https://github.com/buildkite/buildkite/pull/32685) makes pipelines connected to the GitHub Code Access App persist native release builds at the tag's immutable peeled commit. The buildkite-gha docs still described checked-out `HEAD` normalization as the primary release identity.

## What

Document the server-resolved SHA path, the Code Access App requirement for immutable provenance and hosted release `GITHUB_TOKEN` issuance, and the local `HEAD` compatibility fallback for other pipelines.

Extend the existing release plugin integration test to prove full SHAs bypass commit normalization and flow unchanged into event plans and workflow-token policy, while preserving fallback coverage.


sj26:

PR #347, opened 2026-08-20
PB-3032: Snapshot checkout contracts per commit

zhming0:
## Why

Workflows pinned to historical `actions/checkout` commits fail when those commits fall outside the current release and `main` snapshots. Blanket future admission is compatibility-unsafe even though the native adapter never executes upstream code: declared inputs, defaults, and outputs changed between commits.

Closes [PB-3032](https://linear.app/buildkite/issue/PB-3032/admit-frozen-actionscheckout-history-with-per-commit-contracts).

## What

- Snapshot immutable commits reachable from upstream tags, `main`, `master`, and `releases/v1` through `releases/v6`, covering 254 SHAs across 16 contracts—159 more SHAs than before.
- Validate explicit inputs and preserve full-history defaults per commit, and expose `ref` and `commit` outputs only when the manifest declares both. This corrects the previous pre-v4.2 output classification.
- Keep unknown and future SHAs fail-closed until the snapshots are deliberately regenerated.

`mise run check` passes, including tests, race tests, lint, vet, local smoke checks, and release validation.


sj26:
Released in [v0.37.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.37.0)

PR #352, opened 2026-08-21
Run DeepSec directly with npx

wolfeidau:
## Why

The DeepSec workspace only needs the CLI, so maintaining a separate npm manifest and lockfile adds unnecessary dependency metadata.

## What

Run the pinned DeepSec 2.3.6 CLI directly through `npx`, and keep the TypeScript configuration dependency-free so it resolves without a local `node_modules`. Remove the obsolete npm manifest and lockfile and update the workspace documentation.


lox (approved):

sj26:

PR #31, opened 2026-07-30
Publish job summaries as Buildkite annotations

mcncl:
## Summary

- publish non-empty, bounded and scrubbed job summaries as job-scoped Buildkite annotations
- use a stable context so retries update rather than duplicate the annotation
- keep annotation failures advisory and preserve the authoritative terminal result manifest
- document the Buildkite Agent v3.112+ requirement and summary size behavior

## Contract

The runtime invokes `buildkite-agent annotate` with Markdown on stdin, `--scope job`, the validated Buildkite job UUID, context `buildkite-gha-job-summary`, and `info` style. Publication happens only after the authoritative result artifact and metadata mirrors have been attempted. Empty summaries are a no-op; summaries from failed and cancelled jobs are still eligible for publication.

## Verification

- `mise run check`
- focused transport/runtime/CLI tests cover argument and stdin shape, validation, stable ordering, advisory failure, empty summaries, failed/cancelled jobs, and malformed-result fallback

PR #355, opened 2026-08-21
Model first-party pull request synchronization builds

sj26:
## Why

Buildkite's GitHub integration coalesces first-party `pull_request.synchronize` deliveries into linked push builds enriched with pull request metadata. buildkite-gha treated `BUILDKITE_GITHUB_EVENT=push` as authoritative, so a workflow with filtered `push` and `pull_request` triggers selected the push context and rejected the feature branch. Rebuilds without linked push metadata selected the pull request context instead, producing inconsistent results.

This was reproduced by sj26/mailcatcher Buildkite build #31.

## What

- Model a GitHub push with a positive Buildkite pull request number as one `pull_request`/`synchronize` event using trusted Buildkite head and base metadata.
- Keep ordinary branch pushes as `push`, partition generated conditions so the event groups cannot overlap, and preserve Buildkite repository, commit, and checkout authority.
- Treat the linked push payload as insufficient pull request path-filter evidence, so those filters fail closed.
- Cover linked-webhook and rebuild paths with focused event, trigger, path-filter, and plugin integration tests, and document the compatibility and security boundaries.


sj26:
I don't think we need to do this. With the GitHub Actions pipeline trigger and the github checkout action I think it'll all act roughly as it should without any change.

PR #360, opened 2026-08-21
Support compile-time job container image expressions

lox:
## Why

Job container images currently reject expressions even when their values are immutable and fully known during compilation. This blocks otherwise compatible workflows that select images from static matrix entries, reusable inputs, variables, or GitHub event metadata.

## What

Resolve job container image expressions for each statically expanded job instance using the existing compile-time expression context. Complete expressions must produce a non-empty string and the final value must pass the existing image validation; secrets, runtime outputs, whole contexts, and deferred expressions remain unsupported.

Plans stay literal-only, so image selection remains deterministic and the existing runner, registry credential, platform, and runtime policies are unchanged. Compatibility documentation and focused parser, expression, compiler, plan, pipeline, and runtime coverage are included.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -551,6 +551,7 @@ services:
 
 Services support `image`, `credentials`, `env`, `ports`, `volumes`, `options`, `command`, and `entrypoint`.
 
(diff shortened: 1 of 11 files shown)


sj26:
Released in [v0.31.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.31.0)

PR #365, opened 2026-08-22
Make GitHub token checks follow expression behavior

lox:
## Why

The compiler must decide whether a job may receive `github.token` before the job runs. Today, token planning and runtime execution interpret expressions through separate code paths. That means support for operators and functions can work at runtime but still need another field-specific planning fix.

For example:

```yaml
${{ github.server_url == 'https://github.com' && github.token || '' }}
```

Both paths need to agree that:

- GitHub requires the token.
- Origin skips the token branch and must not receive token access.
- An unknown runtime condition must conservatively allow token access if that branch could run.

Maintaining those rules twice caused the repeated reachability fixes discussed in [the expression authority design](https://github.com/buildkite/buildkite-gha/blob/main/docs/plans/expression-authority-planning.md).

## What

Make runtime evaluation and token planning share one implementation of expression traversal, short-circuiting, comparisons, and function argument selection.

Runtime evaluation still produces concrete values. Planning now evaluates the same expression with values that may be known or unknown and records whether a reachable branch needs `github.token`. Validation remains a separate exhaustive pass, so invalid or prohibited references are still rejected even in branches that would not run.

This is the first, behavior-preserving slice of the design. It moves action input defaults onto the shared evaluator; later slices will normalize workflow fields and resolved actions so new expression-bearing fields no longer require separate authority scanners.

### DIFF
--- a/internal/expression/abstract.go
+++ b/internal/expression/abstract.go
@@ -1,0 +1,143 @@
+package expression
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+
+	"github.com/rhysd/actionlint"
+)
+
+// AbstractValue is either one concrete expression value or an unknown runtime
+// value. Unknown is distinct from a known null value.
+type AbstractValue struct {
+	Known bool
+	Value any
+}
+
+// GitHubTokenEffect records why evaluation can require github.token.
+type GitHubTokenEffect uint8
+
+const (
+	GitHubTokenDirect GitHubTokenEffect = 1 << iota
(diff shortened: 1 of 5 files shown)


lox:
@buildsworth-bk review

lox:
Acknowledged for review 4998578016: no inline issues; awaiting the required human review.

wolfeidau (approved):
👍🏻

sj26:

PR #372, opened 2026-08-22
Allow actions when no GitHub check run ID exists

lox:
## Why

CodeQL v4 asks GitHub Actions for the current check run ID:

```yaml
check-run-id:
  default: ${{ job.check_run_id }}
```

Buildkite does not create a GitHub check run, so there is no ID to return. We currently reject the action, even though CodeQL treats a missing ID as optional and continues without it.

## What

Treat the missing ID the same way GitHub does when it is unavailable:

```text
${{ job.check_run_id }} → ""
```

This only applies to action input defaults. We do not invent an ID, expose other `job` fields, or request a GitHub token for code that cannot run because the ID is empty.


wolfeidau (approved):
👍🏻

sj26:

PR #374, opened 2026-08-22
Make pull requests easier to understand

lox:
## Why

A technically correct pull request can still make reviewers decode the implementation before they understand the problem:

| Avoid | Prefer |
| --- | --- |
| `Support unavailable action check run IDs` | `Allow actions when no GitHub check run ID exists` |

## What

Add repository guidance for plain-language titles and short `Why`/`What` descriptions. Prefer a concrete failing example and the new result over subsystem terminology or another explanatory paragraph.

Do not force examples for internal refactors, and omit routine test output and file inventories.


sj26:
Released in [v0.33.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.33.0)

PR #389, opened 2026-08-25
PB-3075: Explain unsupported workflow file paths

wolfeidau:
## Why

An unsupported workflow path did not explain all likely causes or how to correct the configuration. Fixes PB-3075.

## What

The error now tells users to check that the workflow exists at the configured path and is not a symlink, and lists symlinks, untracked files, directories, and globs as unsupported. Missing configured workflows retain their existing skip behavior.


sj26:
Released in [v0.35.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.35.0)

PR #390, opened 2026-08-25
Upload importer artifacts in one batch

wolfeidau:
## Why

The importer currently starts one `buildkite-agent artifact upload` process for every executable, plan, and generated failure artifact. A workflow with many jobs therefore starts many agent processes before it can upload its pipeline.

## What

Materialize and verify every importer artifact under the private artifact root, then upload them together:

```sh
buildkite-agent artifact upload ".buildkite-gha/**/*" --concurrency 8
```

The pipeline remains last and is not uploaded after artifact failure, cancellation, or an integrity failure. This path now requires Buildkite agent v3.129 or newer.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -295,7 +295,7 @@ buildkite-gha compile \
 buildkite-gha upload .github/workflows/ci.yml
 ```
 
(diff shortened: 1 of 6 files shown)


sj26:
Released in [v0.35.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.35.0)

PR #396, opened 2026-08-25
Anchor hosted cache validation to a direct mount

lox:
## Why

[PR #392](https://github.com/buildkite/buildkite-gha/pull/392) assumed `/cache/bkcache` was the mounted cache-volume root. The live v0.35.1 proof in [buildkite-gha-benchmarks build #68](https://buildkite.com/buildkite/buildkite-gha-benchmarks/builds/68) disproved that: Hosted mounted `/dev/vdi` directly at the two configured Gradle paths and the managed mise path, while `/cache/bkcache` itself was not a mountpoint. Bootstrap therefore failed before `run-job`.

## What

Use a compiler-owned exact cache path as the filesystem anchor. Jobs that need mise use the managed mise path already added to the volume. Configured-only Linux runner-user jobs add a private validation path to the same volume.

The anchor must be a direct mountpoint, or a descendant of a verified legacy `/cache/bkcache` mount. Each configured path must be an available directory on the anchor's filesystem device and either an exact mountpoint or a descendant of that verified legacy root. This accepts Hosted's per-path direct mounts while rejecting ordinary directories, unrelated mounts, unavailable paths, and the entire cache root. `BUILDKITE_AGENT_CACHE_PATHS` is not authority, and recursive ownership remains limited to configured targets.

Checks passed:

- `go test ./internal/buildkite ./internal/cli`
- `mise run check`
- `git diff --check`

This requires another patch release followed by the same live benchmark proof. This PR does not release or run that proof.

### DIFF
--- a/internal/buildkite/experimental_runner_user.go
+++ b/internal/buildkite/experimental_runner_user.go
@@ -28,7 +28,11 @@ func experimentalRunnerUserBootstrap(requiresMise, hostedToolCache bool, cache *
 		`sudo -n --user runner -- test -r "$plan" && ! sudo -n --user runner -- test -w "$plan" || { echo 'buildkite-gha: runner plan permissions are unsafe' >&2; exit 1; }`,
 	}
 	if cache != nil {
-		commands = append(commands, experimentalRunnerCacheOwnershipCommands("/cache/bkcache", cache.Paths)...)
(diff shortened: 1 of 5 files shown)


sj26:
Released in [v0.36.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.36.0)

PR #403, opened 2026-08-26
PB-3103: Ignore merge_group path filters like GitHub does

zhming0:
## Why

Workflows that put path filters on the merge queue trigger are rejected:

```yaml
on:
  merge_group:
    paths: ["src/**"]
```

```
E_PIPELINE_GENERATION: Merge_group trigger path filters cannot be translated safely.
Remove paths and paths-ignore from this trigger, or move the filtering into a job or step.
```

GitHub itself does not evaluate `paths` or `paths-ignore` for `merge_group` events — every merge queue delivery runs the workflow regardless ([community discussion #45899](https://github.com/orgs/community/discussions/45899)). Rejecting them made us stricter than GitHub and blocked otherwise-compatible mono-repo workflows.

## What

The workflow above now compiles. The filters are ignored, matching GitHub's delivered behavior, and a warning tells the author:

```
W_MERGE_GROUP_PATH_FILTERS_IGNORED: on.merge_group paths and paths-ignore are ignored,
matching GitHub, which does not evaluate path filters for merge_group events. Every
merge_group delivery runs this workflow. Move the filtering into a job or step condition
if you need it.
```

Tag and workflow filters on `merge_group` remain rejected. Clears the `merge-group-filters` compatibility gap (`mise run compatibility:gaps` now reports it passing).

Closes PB-3103.


sj26:
Released in [v0.40.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.40.0)

PR #407, opened 2026-08-28
Compile unknown download-artifact commits with a warning

zhming0:
## Why

A moving or newly pinned `actions/download-artifact` ref can resolve to a valid immutable SHA outside the exact admission set and fail compilation, even though buildkite-gha replaces the action with its producer-bound native adapter.

## What

Unknown lowercase 40-hex commits now use the stable v8.0.1 contract and emit one source-located warning per distinct SHA. Known commits retain their exact contracts, and malformed identities still fail. Downloads remain confined to artifacts from verified direct `needs` producers with the existing selector, destination, ZIP, digest, count, and size bounds; upstream JavaScript never runs.

Closes PB-3113.


zhming0:
Superseded by #408, which combines the upload and download fallbacks into one commit on main.

PR #408, opened 2026-08-28
Compile unknown artifact commits with warnings

zhming0:
## Why

Moving or newly pinned `actions/upload-artifact` and `actions/download-artifact` refs can resolve to valid immutable SHAs outside their exact admission sets and fail compilation, even though buildkite-gha replaces both actions with bounded native adapters.

## What

Unknown lowercase 40-hex upload commits now use the stable v7.0.1 contract; unknown download commits use v8.0.1. Each adapter emits one source-located warning per distinct SHA. Known commits retain their exact release contracts, known unsupported upload v3.2.2 remains rejected, and malformed identities fail.

Native upload path, mode, size, and output bounds remain unchanged. Downloads remain confined to artifacts from verified direct `needs` producers with existing selector, destination, ZIP, digest, count, and size bounds. Neither adapter runs upstream JavaScript.

Closes PB-3112 and PB-3113. Supersedes #406 and #407 so both changes reach `main` as one commit.


sj26:
Released in [v0.42.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.42.0)

PR #409, opened 2026-08-28
Run workflows selected by GitHub Actions Pipeline Triggers

sj26:
## Why

GitHub Actions Pipeline Triggers are in private preview. When enabled, they select the workflow as they create a build, so the pipeline should not need to repeat that selector:

```yaml
steps:
  - label: ":github:"
    plugin: github-actions
```

This server-generated importer can also omit a step key. Today, the plugin rejects both the empty configuration and the empty `BUILDKITE_STEP_KEY`, so the Buildkite-selected workflow never runs.

The Buildkite backend contract is shipped. The wrapper plugin still needs a compatible `buildkite-gha` runtime release before promoting this configuration; this PR does not release either component.

## What

- Accept an empty plugin configuration only when `BUILDKITE_GITHUB_WORKFLOW_PATH` marks a Pipeline Trigger selection. Explicit selectors still win, and ordinary empty configurations still fail before side effects.
- Prefer and validate `GITHUB_EVENT_NAME`, `GITHUB_WORKFLOW`, `GITHUB_WORKFLOW_REF`, and `GITHUB_WORKFLOW_SHA` against the selected repository, event ref, workflow file, and normalized checkout commit. Buildkite-prefixed values remain compatibility fallbacks.
- Permit only server-selected importers to omit a step key. Artifacts upload before the dynamic pipeline, remain scoped to `BUILDKITE_JOB_ID`, and generated work omits the importer dependency instead of using a job UUID as a step key. Explicit-selector dependencies are unchanged.
- Derive and protect `github.workflow_ref`, `github.workflow_sha`, `GITHUB_WORKFLOW_REF`, and `GITHUB_WORKFLOW_SHA` from immutable job plans. Jobs expanded from local or public reusable workflows retain the top-level caller identity, including the pull request merge ref with the checked-out head SHA.


sj26:
Released in [v0.48.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.48.0)

PR #410, opened 2026-08-28
PB-3123: Support ${{ github.workspace }} and run identity expressions

zhming0:
## Why

Workflows that read the workspace path or run identity through expressions fail at job setup:

```yaml
env:
  FAKE_HOME: ${{ github.workspace }}/fake-home
  CACHE_KEY: cache-${{ github.run_id }}-${{ github.run_attempt }}
```

```
buildkite-gha: run-job: evaluate job environment: evaluate "FAKE_HOME": unsupported runtime github reference "github.workspace"
```

These are common in migrated workflows: `github.workspace` for paths, and run identity for unique cache keys and artifact names.

## What

The expressions above now resolve, and the matching environment variables are set:

| Expression | Value |
| --- | --- |
| `github.workspace` | The workspace directory: the fixed job-container mount for container jobs, the host checkout directory otherwise. |
| `github.run_id` | `BUILDKITE_BUILD_ID` |
| `github.run_number` | `BUILDKITE_BUILD_NUMBER` |
| `github.run_attempt` | `BUILDKITE_RETRY_COUNT` plus one, matching GitHub's 1-based attempts. |

`GITHUB_RUN_ID`, `GITHUB_RUN_NUMBER`, and `GITHUB_RUN_ATTEMPT` are exposed to steps and protected from `GITHUB_ENV` spoofing like the other identity variables.

Boundaries:

- Referencing run identity outside a Buildkite build is an error rather than an empty string, so derived cache keys cannot silently collide.
- GitHub run URLs or API calls built from `run_id` do not resolve, because no GitHub Actions run exists.
- Job and step `if:` conditions and compile-time positions still reject these fields.

Also moves the test-only `startJobContainer` helper into the test file: golangci-lint's `unused` check fails on macOS because the linux/amd64 test file that calls it is excluded from analysis.

Linear: [PB-3123](https://linear.app/buildkite/issue/PB-3123/support-githubworkspace-githubrun-id-githubrun-number-and-githubrun)

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -933,6 +933,8 @@ The runtime retains this bounded `github` context:
 | `ref_type` | `branch` for branch and pull request refs; `tag` for tag refs. |
 | `action_path` | Composite action directory inside composite steps; empty elsewhere. |
 | `action_repository`, `action_ref` | Remote composite repository and requested ref; empty for local composites and outside composite steps. |
+| `workspace` | The workspace directory: the fixed job-container mount for container jobs, the host checkout directory otherwise. Exposed as `GITHUB_WORKSPACE`. |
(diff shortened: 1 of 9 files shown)


sj26:
Released in [v0.43.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.43.0)

PR #416, opened 2026-08-31
Report the compatibility rejection message whole in failure telemetry

zhming0:
## Why

When a workflow is rejected as incompatible, the failure telemetry event should carry the rejection message. Today it often does not. Sixteen compatibility failures in the last 30 days were unreadable.

The event's `error_message` is the final 1,024 bytes of the command's error output. The rejection line is printed early, and the annotation and pipeline-upload output that follows it fills the 1,024-byte window, so the message that names the rejected feature is pushed out before the event is sent:

```
# error_message today: the tail is agent output, the rejection is gone
…annotation and pipeline-upload output… buildkite-gha: upload: workflow can not be uploaded
```

This is not fixed by raising or removing the 1,024-byte limit. The backend applies the same bound when it logs the message, and a larger window still fills with whatever is printed last.

## What

When a workflow diagnostic already attributes the failure — the same rule that sets `failure_code` — send that diagnostic's message and detail as `error_message`, whole:

```
# error_message now
GitHub environments and environment secrets are unsupported. Remove the environment key from job "deploy". …
```

A message longer than 1,024 bytes keeps its start, which names the rejected feature, and sets `error_message_truncated`. Failures without an attributing diagnostic keep the current behavior: the final 1,024 bytes of error output.

Fixes [PB-3141](https://linear.app/buildkite/issue/PB-3141/fix-compatibility-error-message-truncation-in-buildkite-gha-logs).

`mise run check` passes. One pre-existing test failure in sandboxes with global git commit signing (`TestRunUploadAppliesPullRequestPathFiltersFromGitDiff`) reproduces on `main` and is unrelated; it passes with `GIT_CONFIG_GLOBAL=/dev/null`.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -524,9 +524,12 @@ The plugin value must be a YAML boolean, not a quoted string.
 In Buildkite jobs, the importer and runtime send best-effort completion
 telemetry through the job-authenticated Agent API. Events contain the command,
 outcome, client version, duration, and bounded diagnostic codes and severities.
-For an unsuccessful command, they also contain the final 1,024 bytes of
-normalized user-visible error output. `error_message_truncated` says whether
-earlier output was omitted.
+For an unsuccessful command, they also contain a normalized user-visible error
+message of at most 1,024 bytes. When a workflow diagnostic attributes the
+failure, the message is that diagnostic's text, kept whole so later job output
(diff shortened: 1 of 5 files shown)


sj26:
Released in [v0.46.1](https://github.com/buildkite/buildkite-gha/releases/tag/v0.46.1)

PR #421, opened 2026-09-02
Keep workflow processing stages consistent

zhming0:
## Why

Compiler findings, compatibility reports, and command telemetry share the same workflow-processing stages, but stage IDs, display names, order, and diagnostic codes were owned in different packages. Adding or changing a stage required synchronized string edits across those areas. This is part of PB-3178.

## What

- Add a neutral `workflowprocessing` contract with typed stage IDs, canonical stage metadata, processing order, and diagnostic codes.
- Generate compatibility reports from that registry and carry the stage type through report attribution and telemetry.
- Keep compiler aliases during the migration so existing compiler call sites retain the same behavior while compatibility and CLI consumers move to the shared contract.
- Separate compiler diagnostic attribution from partial processing evidence so filenames identify each responsibility.
- Verify registry integrity and that reports contain the complete canonical stage sequence.

## Verification

`mise run check`

### DIFF
--- a/internal/cli/cli_test.go
+++ b/internal/cli/cli_test.go
@@ -22,12 +22,12 @@ import (
 
 // Stable stage IDs asserted throughout the report expectations below.
 const (
-	stageWorkflowParsing = string(compiler.StageWorkflowParsing)
-	stageEventValidation = string(compiler.StageEventValidation)
-	stageGraph           = string(compiler.StageGraph)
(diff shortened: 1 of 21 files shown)


sj26:
Released in [v0.48.2](https://github.com/buildkite/buildkite-gha/releases/tag/v0.48.2)

PR #423, opened 2026-09-02
Keep Agent API clients consistent

zhming0:
## Why

Cache credentials, GitHub tokens, OIDC tokens, and runner resolution all connect to the job-scoped Buildkite Agent API, but each client repeated the same endpoint and credential validation, HTTP hardening, authentication, and user-agent setup. Changes to this shared policy had to be synchronized across four service implementations. This is part of PB-3178.

Central ownership reduces drift in security-sensitive defaults and gives future Agent API integrations one established boundary to reuse across the repository.

## What

- Add a dedicated `internal/agentapi` package for safe job URL construction, job and token validation, timeout and redirect policy, and standard request headers.
- Keep `internal/transport` focused on artifact and result transport over public `buildkite-agent` commands rather than mixing in the live HTTP API boundary.
- Reuse the client from cache credentials, GitHub tokens, OIDC tokens, and runner resolution while leaving each service's payloads, responses, errors, and public configuration unchanged.
- Cover the shared boundary's authenticated request construction and unsafe configuration rejection directly.

## Verification

`GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=commit.gpgsign GIT_CONFIG_VALUE_0=false mise run check`

### DIFF
--- a/internal/agentapi/client.go
+++ b/internal/agentapi/client.go
@@ -1,0 +1,128 @@
+// Package agentapi provides the authenticated HTTP boundary to the
+// job-scoped Buildkite Agent API.
+package agentapi
+
+import (
+	"fmt"
+	"net"
+	"net/http"
+	"net/url"
+	"path"
+	"strings"
+	"time"
+
+	"github.com/buildkite/buildkite-gha/internal/useragent"
+)
+
+// Config carries the current job's Agent API connection and authentication
+// material.
+type Config struct {
+	Endpoint      string
+	JobID         string
+	JobToken      string
+	ClientVersion string
+	HTTPClient    *http.Client
+}
+
+// Client sends authenticated requests to one job-scoped Agent API.
+type Client struct {
+	jobURL    *url.URL
+	jobToken  string
+	userAgent string
+	client    *http.Client
+}
+
+// New constructs a job-scoped client. The service name identifies the calling
+// integration in configuration errors.
+func New(config Config, service string) (*Client, error) {
+	if !ValidJobID(config.JobID) {
+		return nil, fmt.Errorf("%s Agent job ID is required", service)
+	}
+	jobURL, err := url.Parse(config.Endpoint)
+	if err != nil || !validEndpoint(jobURL) {
+		return nil, fmt.Errorf("safe %s Agent endpoint using HTTPS or loopback HTTP is required", service)
+	}
+	if config.JobToken == "" || strings.ContainsAny(config.JobToken, "\r\n") {
+		return nil, fmt.Errorf("%s Agent job token is required", service)
+	}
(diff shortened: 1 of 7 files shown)


zhming0:
Good question—the runtime-local placement was not justified, so I moved the shared client to `internal/agentapi`.

I did not put it in `transport` because that package's existing contract is narrower: it owns artifact/result transport and invokes public `buildkite-agent` commands through a command runner. Its package documentation explicitly says it avoids the Buildkite API so that boundary can be tested without a live build. Adding authenticated HTTP connection policy there would mix two transport mechanisms and weaken that cohesion.

`agentapi` is an idiomatic single-word Go package name and gives cache, GitHub token, OIDC, runner-resolution, and future Agent API consumers a neutral dependency. Service-specific request/response logic remains in `runtime`; only shared job identity, endpoint construction, authentication headers, timeout, and redirect policy live in `agentapi`. The shared Buildkite job-ID validation is also used by repository-provider credentials.

Relevant package tests and the full `mise run check` gate pass after the move.

wolfeidau (approved):
👍🏻 🚢

internal/agentapi/client.go
```diff
@@ -0,0 +1,113 @@
...
+	return endpoint.String()
+}
+
+// Do adds the Agent API authentication and standard request headers before
+// sending request with the bounded HTTP client.
+func (c *Client) Do(request *http.Request) (*http.Response, error) {
+	request.Header.Set("Authorization", "Token "+c.jobToken)
+	request.Header.Set("Accept", "application/json")
+	request.Header.Set("User-Agent", c.userAgent)
+	return c.client.Do(request)
```
sj26, line 83: `Do` attaches the job token to any supplied request without checking that its scheme, host, and path belong to this client’s configured job endpoint. A focused `httptest` with the client configured for server A and a request targeting server B resulted in B receiving `Authorization: Token job-secret`.

The current four callers build fixed URLs with `agent.URL`, so this is not presently driven by untrusted input. However, this package is intended as the reusable security boundary for future Agent API integrations, and the API turns a caller mistake or later untrusted URL flow into job-token exfiltration. Could `Do` reject requests outside `c.jobURL`’s origin and job-path prefix, or could the client own request construction so arbitrary targets cannot receive credentials?

zhming0: Fixed in [`b3493b8`](https://github.com/buildkite/buildkite-gha/commit/b3493b8): `Do` now rejects requests before attaching credentials unless their scheme and host match the configured Agent API and their cleaned path remains at or below the configured job endpoint. This also rejects an overridden Host header. The regression test covers a different origin, sibling job, parent path, and path traversal, and proves no request reaches either server. Relevant package tests and the full `mise run check` gate pass.

sj26:
Released in [v0.48.4](https://github.com/buildkite/buildkite-gha/releases/tag/v0.48.4)

PR #427, opened 2026-09-02
Prevent agent test failures and clarify architecture ownership

zhming0:
## Why

Orb tests that create temporary Git repositories inherit required commit signing without an available key, so `mise run check` fails with `No signing key is available for this commit`. Recent work also repeatedly needed corrections for generic package names, duplicated planning/runtime semantics, and Buildkite server contracts that were still changing.

## What

Orb setup now disables commit signing in its global Git configuration so temporary repositories work without per-command overrides. `AGENTS.md` adds three focused architecture prompts: use package docs to place and name code, consult the expression-authority contract before semantic changes, and confirm server contracts and rollout order before implementing CLI support. This is part of PB-3178.

### DIFF
--- a/.agents/setup
+++ b/.agents/setup
@@ -50,6 +50,10 @@ if ! command -v gh >/dev/null; then
   exit 1
 fi
 
+# Orb system configuration requires commit signing but does not provide its
+# signing key. Tests create temporary repositories, so override it globally.
+git config --global commit.gpgsign false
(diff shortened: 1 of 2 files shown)


sj26:
Released in [v0.48.8](https://github.com/buildkite/buildkite-gha/releases/tag/v0.48.8)

PR #431, opened 2026-09-02
Separate runtime tests by behavior

zhming0:
## Why

`internal/runtime/runtime_test.go` mixed Docker execution, process control, output processing, concurrency internals, shared fixture construction, and job orchestration in 8,203 lines. That made ownership unclear and hid the genuinely unique security, ordering, failure, platform, reusable-workflow, and composite-action coverage. This maintenance is part of PB-3178.

## What

Reduce `runtime_test.go` to 5,387 lines (−2,816, 34.3%). Move 64 tests and 2,743 existing lines beside their production responsibilities: Docker actions, process execution, command output processing, direct concurrency internals, files, needs, secrets, workflow binding, and shared runtime fixtures.

Eliminate or consolidate 73 lines of redundant coverage:

- Remove metadata-collision, unsupported-runtime, missing-selector, and duplicate Docker-dispatch assertions where parser, plan-validation, failure-classification, and lifecycle tests prove the same contracts more strongly.
- Run the explicit-cancellation integration once instead of 20 times; the direct supervisor contention test and race suite retain concurrency coverage. This test drops from 0.62s to 0.04s, while total package time remains within run-to-run noise.
- Consolidate the equivalent runner-debug and check-run-ID action-default tests into one table without dropping either case.

The remaining file contains 152 job/action integration tests. It retains unique token and secret isolation, condition and failure ordering, lifecycle cleanup, platform environment, reusable-workflow, remote-action, and nested-composite behavior. PR #434 changes that orchestration contract, so this PR does not depend on or pre-emptively rearrange its unmerged semantics.


sj26:
Released in [v0.48.12](https://github.com/buildkite/buildkite-gha/releases/tag/v0.48.12)

PR #432, opened 2026-09-02
Separate corpus report tallying from corpus preparation

zhming0:
## Why

The public workflow corpus command's 951-line entry file owned two different domains: preparing and validating the corpus, then interpreting every processing report and presenting aggregate results. Understanding report precedence, diagnostics, and tally JSON required navigating past acquisition, extraction, sampling, and validator setup.

This maintainability work is part of PB-3178.

## What

Move report schema checks, workflow inspection, repository outcome precedence, deterministic summaries, and tally JSON into one focused owner with its focused test.

`main.go` drops from 951 to 647 lines (−304, 32%), and `main_test.go` drops from 392 to 314 lines. The command flow and report interpretation now have separate files while preserving package visibility, call sites, output ordering, error text, and JSON fields.

### DIFF
--- a/cmd/validate-public-workflow-corpus/main.go
+++ b/cmd/validate-public-workflow-corpus/main.go
@@ -13,17 +13,13 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/fs"
 	"os"
 	"os/exec"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


zhming0:
Closing without merge because this command is no longer a PB-3178 priority.

PR #438, opened 2026-09-03
Name runtime output processing explicitly

zhming0:
## Why

`commandProcessor` can read as though it executes commands, but its responsibility is consuming process output: masking logs, interpreting GitHub workflow commands, and collecting Buildkite annotations.

## What

Rename it to `commandOutputProcessor`, including its file, constructor, runtime call sites, and test references. Update the top-level documentation to state that it consumes both process output and runner messages. Behavior is unchanged.

This is part of PB-3178.


sj26:
Released in [v0.48.11](https://github.com/buildkite/buildkite-gha/releases/tag/v0.48.11)

PR #441, opened 2026-09-03
Run pull request workflows without waiting for merge commits

sj26:
## Why

Buildkite builds the pull request head, but importing its workflow currently requires GitHub's temporary merge commit. GitHub can deliver the webhook before that commit exists:

```json
{"mergeable": null, "merge_commit_sha": null}
```

The importer then reports `Pull_request trigger path filters could not be evaluated safely`, even when the workflow has no path filters.

## What

Use the PR head model consistently:

- Without `paths` or `paths-ignore`, skip changed-path evaluation.
- With path filters, compare the pinned base and head commits using a three-dot diff. Require full commit SHAs, complete local history, and one merge base verified as an ancestor of both commits.
- Validate the checkout and filtered workflow against the PR head, not a temporary merge. An advancing base-branch ref does not change the pinned comparison.

Neither `merge_commit_sha` nor `mergeable` participates in path evaluation. This also permits evaluation for conflicting and closed PRs when the head checkout and history are available. Existing diff bounds and rejection of nonmatching filters remain unchanged; this does not switch Buildkite to GitHub Actions' merge-based execution model.


sj26:
Released in [v0.58.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.58.0)

PR #443, opened 2026-09-04
Map UI and API builds to push events

wolfeidau:
## Why

Builds started from the Buildkite UI or API currently select `workflow_dispatch`, so workflows configured only for `push` are skipped.

## What

Map `BUILDKITE_SOURCE=ui` and `BUILDKITE_SOURCE=api` to `push` in both event snapshots and live group conditions. Explicit event snapshots and authoritative `GITHUB_EVENT_NAME` or `BUILDKITE_GITHUB_EVENT` values still take precedence, including rebuild and Pipeline Trigger metadata.

Update skipped-workflow guidance and the documented compatibility contract to match.


zhming0 (approved):

sj26:
Released in [v0.50.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.50.0)

PR #444, opened 2026-09-05
Compile direct and reusable workflows through one job flattener

lox:
## Why

Workflows that call reusable workflows and workflows that do not were flattened into per-job records by two separate code paths. Every field added to that record had to be set in both places, and nothing caught a miss: the blocker-detail provenance added in bd45b5b reached direct workflows only after a same-day follow-up. The recursive path also threaded the caller's state through 15 positional parameters, so the rules for how a called workflow inherits permissions, secrets, call position, guards and concurrency gates lived mid-loop.

## What

Every workflow is now the root of one call tree. A `callFrame` value carries what a workflow inherits from its caller, and its methods own the derivation: the root frame has no permission ceiling and unrestricted secret authority; a child frame narrows to the calling job's permissions, keeps the outermost call position, and extends the namespace and label. Flattened jobs are constructed in one place. The separate direct-workflow path is deleted.

Compiled plans and pipelines are byte-identical for every workflow fixture in the repository across push, dispatch and checkout events at both trust levels. Three diagnostics change for direct workflows because they now share the reusable-workflow path:

- An invalid `needs` graph fails once at the workflow path instead of a raw-path error followed by a `prerequisite "b" has no expanded instances` cascade, and the validation report lists no evaluated jobs.

  ```
  before: /abs/.github/workflows/ci.yml:3:3: job "a": workflow job graph contains a cycle
          ./.github/workflows/ci.yml:3:3: job "a": prerequisite "b" has no expanded instances
  after:  ./.github/workflows/ci.yml:3:3: job "a": workflow job graph contains a cycle
  ```

- The 1,024 job limit that applied after reusable-workflow expansion now applies to every workflow; the error reads `workflow job graph expands beyond 1024 jobs`. `docs/compatibility.md` updated.
- A workflow outside `.github/workflows` that calls a reusable workflow fails at the call site (`ci.yml:4:11: job "delegated": reusable workflows require the caller under .github/workflows`) instead of a path-only error. Direct workflows outside `.github/workflows` still compile.

One subtler change: a workflow that declares `on: workflow_call` and also calls reusable workflows no longer rejects `${{ inputs.* }}` expressions left in its own jobs as "not statically resolvable". The rule is now: callee inputs must resolve statically or defer to a caller need; root inputs are runtime values from the triggering event. The old gate on `on: workflow_call` was an artifact of the split, not a policy: the check and the direct-workflow path were introduced together in a5a0928 (#2), so callable roots without calls have skipped it since day one, and the check's error text is about callee inputs.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -1505,7 +1505,7 @@ hosted-toolchains images provide. macOS images are unsupported.
 | Matrix instances per job | 256 |
 | Reusable workflow nesting | 4 levels |
 | Workflow file | 1 MiB |
-| Jobs after reusable workflow expansion | 1,024 |
+| Jobs in a workflow after reusable workflow expansion | 1,024 |
 | Nested local action depth | 10 levels |
 | Background steps active at once | 10 |
 | Job outputs | 64 |
--- a/internal/compiler/reusable.go
+++ b/internal/compiler/reusable.go
@@ -94,7 +94,6 @@ type reusableResolution struct {
 type reusableResolver struct {
 	workspaceRoot         string
 	repositorySource      RepositorySource
-	stack                 []reusableSourceIdentity
 	materialized          []actionsource.Materialized
 	context               expression.CompileContext
 	rootPermissions       *workflow.Permissions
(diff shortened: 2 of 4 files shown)


lox:
🤖 Acknowledged for review 5119080584: no implementation findings; approval deferred to a human reviewer as noted. Head d7b2742 is unchanged and all checks pass.

PR #449, opened 2026-09-07
PB-3184: Allow needs outputs inside larger reusable workflow input values

zhming0:
## Why

A reusable-workflow call that embeds a job output in a larger input value failed to compile:

```yaml
jobs:
  meta:
    runs-on: ubuntu-latest
    outputs:
      tag: ${{ steps.meta.outputs.tag }}
    steps: [...]
  build-image:
    needs: meta
    uses: ./.github/workflows/build.yml
    with:
      tags: |
        type=raw,value=${{ needs.meta.outputs.tag }}
      flavor: ${{ format('latest={0}', needs.meta.outputs.flavor) }}
```

```
Reusable workflow input "tags" uses a needs expression in an unsupported form.
Pass the whole value as exactly ${{ needs.<job>.outputs.<name> }}, with nothing around it. ...
```

Only the exact whole-value form `${{ needs.<job>.outputs.<name> }}` was accepted.

## What

The call above now compiles. Buildkite resolves the referenced outputs from the verified producer results and renders the value before the called job runs, so the callee receives `type=raw,value=v4.3.0`.

Inside a needs-dependent input, everything other than the `needs.<job>.outputs.<name>` references must resolve before jobs run (literals, graph-time `github`, `vars`, matrix values, static parent inputs, supported operators and pure functions). One value can read several outputs from several jobs. Each referenced job must still be listed in the call's `needs`, and only `string` inputs can take a needs value.

Still rejected, with a message naming the reason:

| Value | Result |
| --- | --- |
| `prefix-${{ needs.prepare.result }}` | `needs reference "needs.prepare.result" must be needs.<job>.outputs.<name>` |
| `${{ needs.prepare.outputs.tag }}-${{ github.run_id }}` | `compile-time expression references unavailable value "github.run_id"` |
| `prefix-${{ inputs.target }}` where `target` carries a needs value from the parent call | `Forward it as exactly ${{ inputs.target }}, or compute the value where the needs output is referenced directly.` |

### Implementation notes for review

- A deferred input in the job plan is now `{template, need_sources, need_outputs}` instead of a single `{sources, outputs}` projection named `value`. The shape mirrors call guards, and the plan validator checks that the template reads only bound prerequisites and binds only prerequisites it reads. The job-plan schema changes accordingly.
- `internal/expression` gains a `deferred-input` profile (template form, `needs` context only, empty for missing outputs, no token) and `ReduceDeferredInput`, which folds graph-time subtrees and returns the residual template with its `needs` references.
- At runtime the deferred-input template is evaluated against the referenced outputs only; the caller's need results are not exposed to the callee.

Linear: https://linear.app/buildkite/issue/PB-3184


sj26:
Released in [v0.52.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.52.0)

PR #450, opened 2026-09-07
PB-3212: Accept compatible upload-artifact commits across upstream history

zhming0:
## Why

A workflow pinned to `actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08` (`v4.6.0`) can fail native adapter admission even though its action contract is compatible:

```
actions/upload-artifact native adapter does not admit resolved commit "65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08"
```

Admitting that commit alone would repeat the problem for other compatible upstream commits.

## What

Snapshot compatible `actions/upload-artifact` manifests from upstream tags and release branches, following the existing `actions/checkout` approach. Exact commits retain their declared inputs, outputs, hidden-file default, and v1 path behavior. Valid immutable commits outside the snapshot continue to use the bounded v7 contract fallback; malformed SHAs and GHES-only v3 security backports remain rejected.

[PB-3212](https://linear.app/buildkite/issue/PB-3212/snapshot-actionsupload-artifact-contracts-per-commit)


sj26:
Released in [v0.54.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.54.0)

PR #452, opened 2026-09-08
Allow empty activity lists for issues and issue comments

sj26:
## Why

GitHub accepts this workflow trigger and runs it for every issue activity, but buildkite-gha rejects it with `E_WORKFLOW_SYNTAX: "types" section should not be empty`:

```yaml
on:
  issues:
    types: []
```

The same failure affects `issue_comment`. Both exact native-accepted workflows from production verification reproduce the failure in importer 0.55.0.

## What

Treat `types: []` as omitted/default-all for **only** `issues` and `issue_comment`, through parsing, trigger translation, and snapshot matching.

Accept the pinned parser's empty-sequence diagnostic only at verified `types` sequence positions for those events; its model already represents these as omitted. Preserve source positions, malformed/unknown type rejection, other filters, and empty-list rejection for unrelated events. Update compatibility guidance and retain the native fixtures as CLI regressions.

This fixes the importer boundary only. Server matching, releases, and rollout are separate.


sj26:
Local validation and review: `mise run check` passed (builds, full tests and race suite, lint, vet, vulnerability check, local smoke inventory, release-configuration checks). Focused parser/translation/CLI regressions passed, including unknown and malformed values, unrelated empty lists, aliases, source positions, and nonempty match/mismatch controls. Both exact native fixtures now return `compilable`, one job/instance, and zero diagnostics through the executable CLI. Local correctness and simplicity review found no material blocker. No package release, deployment, or shared production test writes were performed.

sj26:
Released in [v0.56.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.56.0)

PR #455, opened 2026-09-09
PB-3237: Accept compatible actions/cache commits across upstream history

zhming0:
## Why

`actions/cache` was admitted only at a handful of hand-listed release commits. Anything else failed at compile time, even commits that run the same cache-v2 client the Buildkite Results service already speaks:

```text
actions/cache@main resolved to commit 3edfce9056124e459a23f683a21433670d47daca, which is not admitted: unsupported actions/cache version for Buildkite cache-v2 service ...
```

The same failure hits every workflow using a floating tag such as `actions/cache@v6` the moment upstream tags a new release, until someone hand-edits the list.

## What

Two changes, matching the checkout (PB-3032) and upload-artifact (PB-3212) approach:

1. **Frozen per-commit snapshot.** A generator (`go generate ./internal/action/integration`) walks every upstream tag plus the `main` and `releases/v5` branches and records each commit whose root, `restore`, and `save` bundles all select cache v2 (`ACTIONS_CACHE_SERVICE_V2`, `ACTIONS_RESULTS_URL`) with one embedded `@actions/cache` client ≥ 4.0.0. 162 commits are captured; 161 are admitted (v3.4.1 is excluded because upstream withdrew it). Snapshot commits run their own upstream bundle exactly as before.

2. **Audited release substitution for anything else.** A ref that resolves outside the snapshot never runs that unknown bundle. The compiler instead fetches and runs the newest audited release for the requested major (`v5` → v5.1.0), or v6.1.0 when the major has none (`v2`, `main`, bare SHAs). The plan keeps the requested ref and records the substitute commit, and compilation warns once per distinct resolved commit:

```text
W_CACHE_UNKNOWN_COMMIT_SUBSTITUTED: actions/cache@v6 resolved to commit <sha>, which is not in the frozen actions/cache snapshot admitted to the Buildkite cache-v2 service. The audited v6.1.0 release (55cc8345863c7cc4c66a329aec7e433d2d1c52a9) runs instead. Pin actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 to remove this warning.
```

| Reference | Before | After |
| --- | --- | --- |
| `actions/cache@55cc83…` (v6.1.0) | runs v6.1.0 | runs v6.1.0 |
| `actions/cache@2c8a9b…` (v6.0.0) | runs v6.0.0 | runs v6.0.0 |
| `actions/cache@3edfce…` (`main` tip) | rejected | runs `main` tip (snapshotted) |
| `actions/cache@v6` after a new upstream v6.2.0 | rejected until regeneration | runs v6.1.0 with `W_CACHE_UNKNOWN_COMMIT_SUBSTITUTED` |
| `actions/cache@v2.1.7` (cache v1) | rejected | runs v6.1.0 with warning |

Boundaries: the substitute must resolve to its recorded commit, and it is re-admitted like any other commit; the hosted upload path re-checks the lock's commit from the artifact, so only audited bundles ever receive the job-scoped cache token. Compatibility docs describe the snapshot, substitution, and warning.


zhming0:
@buildsworth-bk I don't think this is spicy, review again.

zhming0:
@buildsworth-bk I authorized an approval. review

sj26 (approved):
Looks like you know what you're doing! 🚀

wolfeidau (approved):
👍🏻 🚢

sj26:

PR #456, opened 2026-09-09
PB-3245: Allow reusable-workflow calls with if when the called workflow declares concurrency

zhming0:
## Why

A reusable-workflow call with an `if` fails to compile when the called workflow declares `concurrency`, and the whole selected workflow becomes a failing replacement step:

```yaml
# .github/workflows/ci.yml
jobs:
  python-runtime:
    if: github.event_name == 'pull_request'
    uses: ./.github/workflows/build-exe-for-python-sdk.yml

# .github/workflows/build-exe-for-python-sdk.yml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

```
called-workflow concurrency is unsupported for guarded reusable-workflow calls
```

The rejection came from 70435a425f2a4479271c2d758d77cd0f351a5366 together with two cases that really are unsafe: calls with `needs`, and prerequisites that share the called workflow's group (which deadlocks the ordered Buildkite gate). Guards have neither problem. The gate's start and finish steps open and close correctly when the flattened jobs skip at runtime, so keeping the gate is never weaker than GitHub's mutual exclusion.

## What

The call compiles and keeps the called workflow's concurrency gate. The condition decides what happens to the gate:

| Call condition at compile time | Gate | Diagnostic |
| --- | --- | --- |
| True (the example on a pull request) | Emitted, as before for unguarded calls | None |
| False (the example on a push, or `if: false`) | Omitted; the jobs skip without entering the group, as on GitHub | None |
| Runtime-dependent (`vars.DEPLOY == 'true'`) | Emitted; a skipped call still waits for the group and holds it until its jobs finish | New warning `W_REUSABLE_WORKFLOW_CONCURRENCY_ENTERED_BEFORE_CALL_CONDITION` |

Omitting the gate for a condition that is already false matters when the group is shared, such as `deploy-production`: otherwise every push build's skipped jobs would queue behind a running deploy from another build before they could skip.

Dropping the gate with a warning was considered and rejected: it would silently remove mutual exclusion for a call that can run, for example two production deploys at once.

Calls with `needs` stay rejected, including under a false condition, so validation stays exhaustive. Docs: `docs/compatibility.md`.

Linear: PB-3245

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -347,7 +347,7 @@ A top-level workflow that does not declare the effective event is excluded befor
 - Caller-visible aggregate results.
 - Outputs mapped directly from `jobs.<job>.outputs.<name>`.
 - Call-level `if` over caller `github`, `inputs`, direct `needs`, and status functions.
-- Workflow-level concurrency in local and public called workflows. Groups may use the called workflow's static inputs. Each static call-matrix instance gets its own workflow gate.
+- Workflow-level concurrency in local and public called workflows, including behind a call-level `if`. Groups may use the called workflow's static inputs. Each static call-matrix instance gets its own workflow gate. See [Concurrency](#concurrency).
 
 **❌ Unsupported:**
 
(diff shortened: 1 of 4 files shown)


sj26:
Released in [v0.59.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.59.0)

PR #457, opened 2026-09-09
Count issue comment evaluations in public corpus tallies

lox:
## Why

The deterministic 10K public-corpus run classified 152 valid generated `issue_comment` evaluations across 149 repositories as unparseable reports, adding `E_REPORT` findings instead of measuring them.

## What

Accept generated `issue_comment` evaluations when tallying public-corpus processing reports. The tally regression test now verifies the evaluation is counted and the report remains parseable.


sj26:

PR #458, opened 2026-09-09
PB-3246: Resolve compile-time vars references to empty strings when the repository defines no variables

zhming0:
## Why

A `runs-on` that falls back when a repository variable is missing failed to compile inside a Buildkite job, even though the Agent API answered that the repository defines no variables (or that the endpoint is not available yet):

```yaml
runs-on: ${{ vars.DSH_CI_FAILOVER_LINUX || 'dsh-ubuntu-24-04-16core' }}
```

```
E_EXPRESSION_INVALID job "node-24": runs-on expression cannot be resolved at compile time: compile-time expression references unavailable value "vars.dsh_ci_failover_linux"
```

Runtime fields already treated such names as empty strings; compile-time fields did not, which contradicted `docs/compatibility.md`.

## What

When the variables source answered (200 with empty scopes, or 404 because the endpoint is not rolled out), compile-time `vars` references now evaluate to the empty string, so the example above selects `dsh-ubuntu-24-04-16core`. Without a source (`compile` outside a Buildkite job) compile-time `vars` references still fail, as before.

The change distinguishes "no source answered" from "source answered with nothing": `VariableSources` records whether scopes were resolved and hands the compiler an empty map instead of `nil` in the second case.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -722,9 +722,13 @@ combined). Its rejection, rate limit (10 requests per job per hour), or
 GitHub outage fails the compile of every workflow that references `vars` with
 the backend's error and any `Retry-After` delay; other workflows still upload.
 A backend without the endpoint, or an organization that has opted out,
-returns 404, which leaves both scopes empty rather than failing the compile.
-Outside a Buildkite job, `compile` has no variable source, so the scopes are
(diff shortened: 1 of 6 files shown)


sj26:

PR #461, opened 2026-09-09
Avoid outside-root Node failures when running JavaScript actions

sj26:
## Why

sj26 reported [Mailcatcher build 2](https://buildkite.com/sj26-new-1/mailcatcher/builds/2/list?sid=01a08369-6015-46d2-8eaf-0a0ef796f210&tab=output) failing before `ruby/setup-ruby@v1` could run, while the [matching GitHub Actions job](https://github.com/sj26/mailcatcher/actions/runs/33640231144/job/100281159939?pr=593) passed:

```text
cached Node 24 failed validation: validate mise-resolved core:node@24.18.0 installation: path is outside root
refusing to remove Node installation: path is outside root
```

Newer mise versions can reuse a system-wide Node installation even with `--no-config` and a separate `MISE_DATA_DIR`. This reproduces both error lines: the installation falls outside the cache that buildkite-gha validates and may remove. The private build logs were inaccessible, so the failed agent's exact mise version and resolved path remain unconfirmed.

## What

Set `MISE_SYSTEM_DATA_DIR` to the managed cache alongside `MISE_DATA_DIR` for Node installation and lookup. A preinstalled system Node no longer diverts JavaScript actions outside that cache; digest checks, symlink rejection, and deletion boundaries remain unchanged.

The regression covers an existing system Node with an empty managed cache and ensures the system installation stays untouched. Real mise checks confirmed the behavior with 2026.8.14 and compatibility with the minimum 2026.5.12.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -32,6 +32,9 @@ Jobs with JavaScript actions need `mise`. The runtime checks
 `BUILDKITE_GHA_MISE`, then `PATH`, then downloads a verified managed copy.
 Shell-only, native-adapter, and Docker-only jobs do not need it.
 
+When a managed cache is configured, the runtime installs Node there rather
+than reusing system-wide mise installations.
+
 Managed Node binaries require glibc 2.28 or newer. The Go CLI has no glibc requirement.
 
 ## Validate a workflow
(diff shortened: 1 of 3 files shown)


sj26:

PR #465, opened 2026-09-09
Run JavaScript actions when hosted mise overrides its data directory

sj26:
## Why

[Mailcatcher's macOS / Ruby 3.3 job](https://buildkite.com/sj26-new-1/mailcatcher/builds/2/list?sid=01a084ec-71af-4cb5-8160-642606bddcf1&tab=output) still failed with `cached Node 24 failed validation: ... path is outside root` after #461 shipped in v0.60.0. The job's executable SHA-256 exactly matches the published v0.60.0 Darwin binary; it was not running an old release.

A diagnostic on the same hosted `macos-medium` queue found `/opt/mise/bin/mise` is this wrapper:

```sh
#!/bin/zsh --no-rcs --no-globalrcs --errexit
export MISE_DATA_DIR=/opt/mise
exec -a "$(basename ${0:a})" /opt/mise/rust-bin/mise "$@"
```

It overwrites the runtime's managed data directory before calling mise 2026.8.12. Consequently, `mise where core:node@24.18.0` returned `/opt/mise/installs/node/24.18.0`, outside the managed cache. #461 isolated system fallback but did not handle this wrapper.

## What

Pin `MISE_INSTALLS_DIR` directly to the managed cache's `installs` directory for installation and lookup. The wrapper can still set its data directory without redirecting Node. Path containment, digest verification, and refusal to delete outside the cache remain unchanged.

The hosted diagnostic confirmed explicit install-directory isolation returns Node inside `/private/tmp/bkcache/buildkite-gha/mise/darwin-arm64/2026.5.12/installs`, with the expected executable digest. Extend the existing system-install regression to reproduce the wrapper override; it fails with the original error before this fix.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -33,7 +33,8 @@ Jobs with JavaScript actions need `mise`. The runtime checks
 Shell-only, native-adapter, and Docker-only jobs do not need it.
 
 When a managed cache is configured, the runtime installs Node there rather
-than reusing system-wide mise installations.
+than reusing system-wide mise installations. The install directory remains
+pinned even when an agent's mise wrapper overrides `MISE_DATA_DIR`.
 
 Managed Node binaries require glibc 2.28 or newer. The Go CLI has no glibc requirement.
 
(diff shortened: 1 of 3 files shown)


sj26:
Released in [v0.63.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.63.0)

PR #466, opened 2026-09-09
PB-3247: Support runner.environment at runtime

zhming0:
## Why

Workflows that branch on the runner environment fail to compile:

```yaml
if: runner.environment == 'self-hosted'
```

```text
condition reference "runner.environment" is unsupported
```

## What

Support `runner.environment` in valid expression and runtime surfaces and set both it and `RUNNER_ENVIRONMENT` to `self-hosted`.

This PR deliberately leaves the value runtime-dependent during static analysis. It does not prune action preparation or change `github.token` authority decisions; those changes are separated into a follow-up.


sj26:
Released in [v0.62.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.62.0)

PR #467, opened 2026-09-09
PB-3247: Skip actions unreachable on Buildkite agents

zhming0:
## Why

After `runner.environment` is available at runtime, job setup can still prepare an action that cannot run:

```yaml
- if: runner.environment == 'github-hosted'
  uses: owner/action@v1
```

Likewise, `${{ runner.environment == 'github-hosted' && github.token || '' }}` conservatively grants token authority even though Buildkite agents report `self-hosted`.

## What

Treat `runner.environment` as a planning-time constant. Known `github-hosted` branches no longer download or mount action source, pull action images, run `pre` or `post` hooks, or grant unreachable `github.token` authority. Reachable actions and queue-dependent values such as `runner.os` remain conservative.

This follows the runtime support merged in #466.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -1111,6 +1111,11 @@ rooted at `github.event`.
 runners registered outside GitHub, including managed providers. Buildkite
 agents are in the same class whether they use hosted agents or your own
 infrastructure.
+The compiler also treats this value as constant. An action behind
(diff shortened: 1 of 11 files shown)


zhming0:
PR #466 delivered the required runner.environment compatibility fix. This optional planning-time pruning and github.token authority optimization is not being pursued.

PR #472, opened 2026-09-10
fix: make workflow import errors easier to understand

mcncl:
### Context

When a build imports several workflows, an error such as:

```sh
[E_PLAN_CONSTRUCTION] unsupported runtime expression "runner.debug" {job=build.test}
```

leaves users piecing together which workflow failed, where the problem is, and what to do next. Annotations contain more context than logs, but neither consistently points users to the relevant configuration.

### Changes

Bring importer logs and failure-step logs into line with annotations:

- Lead with readable explanations rather than internal error codes, with clearer colours and spacing
- Identify the root workflow and offending source, job, step, and action where known
- Show small excerpts from the configuration actually parsed, and link to the exact GitHub revision and line when verified
- Give steps missing both run and uses concrete examples of valid configuration

Excerpts are deliberately limited to eligible configuration lines to avoid exposing scripts, inputs, or nearby sensitive values. When an excerpt or revision cannot be verified, the error retains its location without inventing source content or a link.

This changes presentation, not which jobs run. It requires no Buildkite server changes; API diagnostic detail and Datadog/PostHog routing remain separate work.

### Relevant links

- [Slop](https://slopcannon.tail952194.ts.net/ben-mcnicholl/pb-3220-error-presentation-investigation/)
- [Linear issue](https://linear.app/buildkite/issue/PB-3220/errors-improvements-in-buildkite-gha)


lox (approved):

sj26:
Released in [v0.64.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.64.0)

PR #473, opened 2026-09-10
PB-3277: Allow reusable workflow concurrency after safe prerequisites

zhming0:
## Why

This call fails whenever `deploy.yml` declares workflow-level `concurrency`, even if `prepare` has no concurrency:

```yaml
jobs:
  prepare:
    runs-on: ubuntu-latest
    steps: [{run: true}]
  deploy:
    needs: prepare
    uses: ./.github/workflows/deploy.yml
```

The error is `called-workflow concurrency is unsupported for reusable-workflow calls with prerequisites`. #456 fixed guarded calls but left this restriction.

## What

Allow calls after prerequisites whose concurrency groups are statically disjoint. The called workflow's opening gate waits for external prerequisites; its closing gate still waits for every member and nested gate. Prerequisite edges allow failure so runtime conditions, including `always()`, retain control over whether jobs run.

Reject matching groups throughout the transitive prerequisite graph, including matrix jobs and nested reusable workflows. Unknown groups and overlaps with enclosing/member groups remain unsupported. Also check emitted dependencies together with ordered concurrency queues: two gates waiting on jobs in each other's groups must fail rather than deadlock.

Buildkite reserves queue positions at upload time, unlike GitHub's admission after prerequisites complete. `W_REUSABLE_WORKFLOW_CONCURRENCY_QUEUED_BEFORE_PREREQUISITES` warns that ready calls or builds can wait behind unfinished prerequisites. Mutual exclusion remains intact; known-false guards still omit their gate. Analysis covers the generated pipeline, not dependencies introduced by other pipelines or later manual uploads.

Closes [PB-3277](https://linear.app/buildkite/issue/PB-3277/allow-reusable-workflows-with-concurrency-after-proven-safe).


zhming0:
@buildsworth-bk I consider this is a bug fixing, we'd allow more workflows that are safe to run to run, not spicy at all. Review again.

wolfeidau (approved):
👍🏻 🚀

zhming0:
Released in [v0.65.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.65.0)

PR #480, opened 2026-09-11
fix: underline offending line and fix line linking

mcncl:
## Description

Errors such as `pull_request_review` has unsupported filters don’t identify which filter failed. Source links can also point to the event declaration rather than the offending field.

## Changes

Identify rejected trigger filters and invalid activity types, link to their configuration, and underline safe excerpts in logs and annotations:

```sh
Error: pull_request_review does not support the branches filter;
remove this filter or move the check into a job or step condition using the event payload

Error source: .github/workflows/pr-reviewed.yml:21:5

> 21 |     branches:
     |     ^^^^^^^^
```

The same underlining applies to eligible `uses`, `runs-on`, and `shell` excerpts. Filter values and other potentially sensitive content remain excluded. Validation behavior is unchanged.


zhming0:
Released in [v0.68.2](https://github.com/buildkite/buildkite-gha/releases/tag/v0.68.2)

PR #484, opened 2026-09-11
PB-3316: Show skipped GitHub Actions jobs in Buildkite

zhming0:
## Why

A job-level `if:` that evaluates to false correctly skips every GitHub Actions step and reports a successful check, but Buildkite shows an ordinary green job:

```text
job · e2e
```

GitHub shows “This check was skipped” for the same result. Without an equivalent signal, a build can look as though its workload ran when it did not.

[PB-3316](https://linear.app/buildkite/issue/PB-3316/show-skipped-github-actions-jobs-in-buildkite)

## What

After publishing the logical `skipped` result, append the skipped state to the Buildkite label:

```text
job · e2e (skipped)
```

The check remains successful and downstream jobs still receive `needs.e2e.result == 'skipped'`. The update is retry-safe and advisory, so presentation cannot change the job result.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -1788,10 +1788,10 @@ hosted-toolchains images provide. macOS images are unsupported.
 
 ### Results, retries, and cancellation
 
-- A runtime-skipped Actions job appears successful in Buildkite while publishing a logical `skipped` result for downstream imported jobs.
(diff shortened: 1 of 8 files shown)


zhming0:
Released in [v0.70.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.70.0)

PR #486, opened 2026-09-11
Support pull_request_target workflows from trusted default-branch source

sj26:
## Why

`on: pull_request_target` cannot currently run through GitHub Actions Pipeline Triggers. Treating it as an ordinary PR event would select attacker-controlled workflow and action source.

## What

Support target imports from the backend-selected immutable default-branch commit, while preserving the original PR payload and head/base context. For a PR targeting `release` in a repository whose default is `main`, workflow source and default checkout use the selected `main` commit; branch filters still match `release`, and paths compare PR base to head.

Require matching server workflow identity, linked payload and clean Git source before compilation. Retain the verified payload for native checkout's fork head/merge SHA guard. Existing token, secret, environment and action-source policies remain in force.

This is bounded support, not full GitHub parity: the backend pins at processing time, path filters require existing local PR history, and target imports reject dirty/sparse/submodule checkouts. PR write tokens and unsafe-checkout/cache-write opt-outs are not added. The compatibility and security guides spell out these limits.

Release this runtime before the companion backend enables target dispatch. The backend must deny target cache writes, including same-repository PRs. Local fixtures exercise distinct default/base/head source and moved refs; no live GitHub or hosted Buildkite end-to-end proof is claimed.

Companion backend: https://github.com/buildkite/buildkite/pull/34172

### DIFF
--- a/cmd/validate-public-workflow-corpus/main.go
+++ b/cmd/validate-public-workflow-corpus/main.go
@@ -35,6 +35,7 @@ const (
 var supportedEvents = map[string]bool{
 	"push":                        true,
 	"pull_request":                true,
+	"pull_request_target":         true,
 	"pull_request_review":         true,
 	"pull_request_review_comment": true,
 	"merge_group":                 true,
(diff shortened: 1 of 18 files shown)


sj26:
I want to be really careful how we introduce support for this event, so flipping this PR back into a draft for now.

PR #491, opened 2026-09-14
PB-3322: Share one prerequisite calculation for job ordering and blocked jobs

zhming0:
Stack: PR 1 of 2. PR 2 (PB-3323, compile request/result interface) is stacked on this branch. Both prepare for the deferred matrix expansion in #468 (PB-3262). Linear: PB-3322.

## Why

Three places in the compiler each work out which earlier jobs a job depends on before it can be compiled: topological ordering, blocked-job classification (a job whose dependency failed evaluation is reported as blocked, not failed), and the reusable-workflow deferred-input walk. They agree today only because each one re-derives the same answer from `needs`, deferred `with:` inputs of a reusable workflow, and the caller `if:` guard of a reusable-workflow call.

#468 adds a fourth consumer: a needs-derived matrix such as

```yaml
build:
  needs: plan
  strategy:
    matrix:
```

is deferred until `plan` finishes, and every job that depends on `build`, directly or through a reusable-workflow input or guard, must be deferred with it. That decision has to use the same dependency list the ordering and blocked-job code use, or a job could be ordered after a dependency the deferral does not see.

## What

`jobPrerequisites` in `internal/compiler/job_graph_expansion.go` is now the one place that answers the question. It returns the sorted, unique set of logical jobs that must exist before a job can be compiled: authored `needs`, the jobs referenced by deferred reusable-workflow inputs, and for each reusable-workflow call the jobs its caller guard and inputs reference. Callee jobs keep their own scope; a caller's `needs` never leaks into the called workflow. Ordering records the result per job and blocked-job classification reads it, replacing `bindingsFailed`, `deferredInputsFailed`, and `deferredInputMembers`.

Behavior is unchanged. Step keys, `needs`, plan bytes and digests, the generated pipeline, diagnostics, and the current unsupported-shape rejections are byte-identical across the 22 local smoke fixtures in both single-workflow and namespaced modes. `internal/compiler/job_prerequisites_test.go` covers the union, deduplication, caller-guard provenance, callee scoping, and blocked-job parity.


zhming0:
Released in [v0.72.1](https://github.com/buildkite/buildkite-gha/releases/tag/v0.72.1)

PR #492, opened 2026-09-14
PB-3323: Compile hosted workflows from one request shared by validation, preflight, and upload

zhming0:
Stack: PR 2 of 2, on top of #491 (PB-3322). Review that one first; this PR's own diff is the last commit. #468 (PB-3262) is the first consumer of the interface introduced here. Linear: PB-3323.

## Why

`upload` compiles each workflow up to four times before it uploads anything: once to validate it against the event, again after the agent resolves runner labels, once more to learn which runtime platforms the jobs need, and finally to produce the pipeline. `validate --profile hosted` compiles twice more. Each pass went through a different `compileHosted*` helper taking 10 to 19 positional arguments, so a pass could silently drop an input the others had. The runtime-platform preflight, for example, compiled without the step-key namespace, OIDC configuration, and event-file flag the final compile used.

#468 needs to run the same compilation again later, inside a deferred step, from inputs the importer recorded, and then prove the result matches what was uploaded. That is only practical when "one hosted compilation of one workflow against one event" is a value that can be built, copied, and compared, not a 19-argument call.

## What

`hostedCompileRequest` (`internal/cli/hosted_compile.go`) holds every input to one hosted compilation: workflow and event bytes, compiler version and distribution digest, importer step, namespace, runner targets and live runner resolution, runtime distribution digests, OIDC, environment source, variables, and the repository source. `validateHostedRequest` and `compileHostedRequest` replace the positional helpers; the compile body is unchanged.

`upload` builds one request per compiled workflow and threads it through validation, runner re-validation, the platform preflight, and the final compile, updating only the field each stage learns (runner resolution, distribution digest, runtime digests, resolved variables). `validate --profile hosted` uses the same request for its validation and preflight. `compiledJobs(bundle)` lists what a compilation produced (step key, logical job, `needs`, plan digest); `upload` uses it to check that action-variable recompilation preserved the job graph, and #468 uses the same shape to record the initial upload and check the deferred recompilation against it.

Generated pipelines, plans, step keys, `needs`, event artifacts, diagnostics, and upload semantics are byte-identical to `main` across the 22 local smoke fixtures in both single-workflow and namespaced modes. Full recompilation still happens where it did; runner policy is still resolved live on every upload. The one intentional difference is the platform preflight, which now compiles with the same namespace, OIDC, and event-file inputs as the final compile; none of them change which platforms a job needs or whether admission passes, but an admission finding from the preflight now names the namespaced step key the final compile would use. `internal/cli/hosted_compile_test.go` checks the request-to-options mapping against expected literals, compiles a real workflow through the request and through the compiler directly and compares plans and pipeline bytes, and covers `compiledJobs` and the job-graph comparison.

Rebasing #468 onto this branch locally: its `hosted.go` changes disappear, its deferred step builds a `hostedCompileRequest` plus two new fields (`RuntimeMatrixRows`, `RuntimeMatrixActionLocks`) instead of assembling options by hand, and the full test suite passes.


zhming0:
Released in [v0.72.1](https://github.com/buildkite/buildkite-gha/releases/tag/v0.72.1)

PR #496, opened 2026-09-14
Reject explicit runner mappings to incompatible queues

jamiemonserrate:
## Why

An explicit runner mapping currently bypasses Agent API resolution:

```yaml
runners:
  - runs-on: ubuntu-latest
    queue: macos-medium
```

If that queue hosts macOS agents, the importer still generates a Linux job and image for it.

## The choice I'm making to fix this

Send each used explicit mapping's queue and inferred platform to the job-scoped Agent API before upload. A hosted OS/architecture mismatch produces `queue_platform_mismatch` diagnostics instead of a runnable child job; an absent queue produces `queue_not_found`.

### This means that missing credentials, an unavailable API, or an older server stop imports using explicit mappings rather than trusting unchecked routing.

Deploy [the companion server support](https://github.com/buildkite/buildkite/pull/34205) before releasing this importer. Customers pinned to older importer versions must upgrade to get this protection.


jamiemonserrate:
CI triage: the shell smoke and public-actions proof in build 2361 both use explicit ubuntu-latest=hosted mappings, but the server returned the automatic-resolution missing_queue error for linux-medium. These live proofs require the companion server support in buildkite/buildkite#34205 before they can pass; do not bypass validation to make them green.

The separate native macOS test failure was a fixture omission: TestPluginRunsNativelyOnDarwinARM64 had no mock queue-validation API. The latest commit adds that fixture without weakening its assertions. Linux CLI tests and lint pass, and the Darwin test binary cross-compiles; native execution remains for CI.

PR #497, opened 2026-09-15
Allow workflow environment fallback expressions

sj26:
## Why

Workflow-level environment values reject fallback expressions that already work in job-level environment values:

```yaml
env:
  CI_TARGET_BRANCH: ${{ github.head_ref || github.ref_name }}
```

## What

Apply the existing job environment expression rules to workflow environment values. This selects the PR head branch for pull requests and the ref name for pushes. Workflow run defaults remain direct-reference-only, and secret inventory and token permission boundaries stay unchanged.

This separates general workflow compatibility from the Windows work in #488. Related to [PB-2723](https://linear.app/buildkite/issue/PB-2723/deferred-platform-gap-add-windows-runner-support); it does not add Windows support.


zhming0:
Released in [v0.75.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.75.0)

PR #498, opened 2026-09-15
Strip shell startup injection from audited cache actions

sj26:
## Why

An audited `actions/cache` process currently inherits values such as `BASH_ENV=/workspace/startup.sh` and `BASH_FUNC_zstd%%` alongside its cache credential. GNU tar may invoke a shell for compression, so a trusted `PATH` alone does not rule out shell startup files or exported functions.

## What

Strip `BASH_ENV`, `ENV`, and all `BASH_FUNC_*` variables before starting audited cache-action processes. Keep the existing trusted `PATH`, other environment filtering, and fresh per-phase credentials unchanged. Other actions retain their normal environment semantics.

Extracted from #488 without its Windows tooling or platform changes. Related to [PB-2723](https://linear.app/buildkite/issue/PB-2723/deferred-platform-gap-add-windows-runner-support); this does not complete Windows support.


sj26:
Closing in favor of GitHub Actions runner compatibility. This is optional cache-specific hardening, not a demonstrated security fix or a prerequisite for Windows support.

PR #499, opened 2026-09-15
Run jobs on the host when container expressions resolve empty

sj26:
## Why

A matrix cannot currently mix Linux host and container jobs:

```yaml
on: push
jobs:
  test:
    strategy:
      matrix:
        target:
          - {name: container, container: 'node:24'}
          - {name: host}
    runs-on: ubuntu-latest
    container: ${{ matrix.target.container }}
    steps: [{run: 'echo container-probe'}]
```

The host entry fails compilation with `expression resolved to <nil>, want string`. An explicitly empty image expression also fails.

## What

Treat null or exactly empty evaluated images as no job container, for both scalar and object forms. The example compiles into one container job and one host job. Enabled containers retain their `env` and `ports`; disabled containers contribute neither settings nor a Docker requirement.

Literal images, whitespace-only results, nonstring results, invalid image references, and forbidden expressions retain their existing validation. Expression contexts and credential authority do not expand.

Extracted from #488 so optional containers can land independently of Windows runner support. This PR adds no Windows support and does not replace #488.

Related: [PB-2723](https://linear.app/buildkite/issue/PB-2723/deferred-platform-gap-add-windows-runner-support); this does not close the Windows platform gap.


zhming0:

PR #500, opened 2026-09-15
Resolve $/ actions and reusable workflows from their containing commit

sj26:
## Why

Workflows such as pytest's call their own actions with GitHub's documented [`$/` self-repository reference](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-using-an-action-in-the-same-repository-as-the-workflow-at-the-running-commit-recommended):

```yaml
steps:
  - uses: $/.github/actions/setup-tox
```

These references currently fail resolution. They select the repository and commit containing the workflow, not the event repository or checkout.

## What

Resolve `$/` actions and [reusable calls](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_iduses) such as `uses: $/.github/workflows/test.yml` at the containing source's immutable commit. Remote composites use their own immediate source and preserve their requested ref in nested `github.action_ref`. Local workflow input must match the fetched bytes at the exact candidate path and commit; synthetic events cannot supply that identity. Authored selectors remain in digest-bound plans, and `./` action behavior is unchanged.

Verified root self/local workflow chains can inherit or map secrets only across declared forwarding edges. Explicit remote calls leave that scope—even at the same repository and commit—and descendants cannot regain it. Private actions remain unsupported; reusable-source access, token aliases, root permissions, and capability planning keep their existing boundaries. Source resolution grants no new credential authority.

Extracted from #488, combining the original self-reference implementation and its secret-forwarding correction. Adapted to main's shared hosted request and existing digest/lock APIs, without Windows support or the optional container/env changes. This is independent of Windows landing.

The pinned pytest source-resolution test retains the expected `github.token` rejection under empty permissions; this does not claim full pytest admission or execution.

The [live lab fixture](https://github.com/sj26-testier/gha-e2e-onboarding-r-fb34348622d8/pull/1) ran at [508f984](https://github.com/buildkite/buildkite-gha/commit/508f9841b53ab1d4beaafc1a123322a978a6c4ff): native GitHub passed 10/10 jobs; Buildkite passed 9/10. After checkout of older source A, `$/` correctly used B on both engines, but Buildkite rejected the subsequent unchanged `./` wrapper because the workspace workflow bytes differed from its plan. That existing restriction remains. Explicit-remote forwarding was not run live; the subsequent symbolic-ref correction has local regression coverage, not a live rerun. `GITHUB_ACTION_REF` is still not populated; commands can receive `github.action_ref` through `env`.

The public `actions/runner` implementation confirms the action-resolution behavior (commit-pinned links):

- [`ActionManager` starts with `WorkflowRepository` and `WorkflowSha`](https://github.com/actions/runner/blob/dde968bf5724d5b9ba7a759f9a114a305e2c76e7/src/Runner.Worker/ActionManager.cs#L189-L201), explicitly including reusable workflows.
- [Nested composite actions use their parent's repository and requested ref](https://github.com/actions/runner/blob/dde968bf5724d5b9ba7a759f9a114a305e2c76e7/src/Runner.Worker/ActionManager.cs#L321-L339).
- [`ResolveSelfRepositoryReferences` rewrites `$/` into that repository/ref](https://github.com/actions/runner/blob/dde968bf5724d5b9ba7a759f9a114a305e2c76e7/src/Runner.Worker/ActionManager.cs#L1548-L1579), and [`ActionRunner` exposes the ref as `github.action_ref`](https://github.com/actions/runner/blob/dde968bf5724d5b9ba7a759f9a114a305e2c76e7/src/Runner.Worker/ActionRunner.cs#L143-L153).

Related: [PB-2723](https://linear.app/buildkite/issue/PB-2723/deferred-platform-gap-add-windows-runner-support) (not resolved by this PR).


wolfeidau (approved):
👍🏻 🚀

zhming0:

PR #501, opened 2026-09-16
Make the default macOS tool cache writable for setup actions

sj26:
## Why

`ruby/setup-ruby@v1` fails on Buildkite macOS agents before tests run:

```text
EACCES: permission denied, mkdir '/Users/runner'
```

Its prebuilt Rubies embed `/Users/runner/hostedtoolcache` and ignore a different `RUNNER_TOOL_CACHE`. Linux already prepares the corresponding `/opt/hostedtoolcache` path.

## What

Prepare `/Users/runner/hostedtoolcache` before macOS jobs run, make it owned and writable by the existing agent user, and select it as `RUNNER_TOOL_CACHE`.

Mac agents now require non-interactive sudo. They keep their existing user and must still provide compatible native dependencies for downloaded tools; Linux behavior is unchanged.


zhming0:

sj26:
It worked 👌

image

PR #502, opened 2026-09-16
PB-3337: Allow jobs to join deferred matrices

zhming0:
## Why

When `build` and `test` take their matrices from job outputs, a downstream job cannot join them. Save this complete workflow as `.github/workflows/deferred-join.yml`:

```yaml
name: Deferred matrix join
on: push

jobs:
  plan-build:
    runs-on: ubuntu-latest
    outputs:
      matrix: ${{ steps.rows.outputs.matrix }}
    steps:
      - id: rows
        run: echo 'matrix=[{"target":"linux"},{"target":"darwin"}]' >> "$GITHUB_OUTPUT"

  plan-test:
    runs-on: ubuntu-latest
    outputs:
      matrix: ${{ steps.rows.outputs.matrix }}
    steps:
      - id: rows
        run: echo 'matrix=[{"suite":"unit"},{"suite":"integration"}]' >> "$GITHUB_OUTPUT"

  build:
    needs: plan-build
    runs-on: ubuntu-latest
    strategy:
      matrix:
        include: ${{ fromJSON(needs.plan-build.outputs.matrix) }}
    steps:
      - run: echo "Building ${{ matrix.target }}"

  test:
    needs: plan-test
    runs-on: ubuntu-latest
    strategy:
      matrix:
        include: ${{ fromJSON(needs.plan-test.outputs.matrix) }}
    steps:
      - run: echo "Testing ${{ matrix.suite }}"

  publish:
    needs: [build, test]
    runs-on: ubuntu-latest
    steps:
      - run: echo "Both matrices finished"
```

Run `buildkite-gha validate .github/workflows/deferred-join.yml`. On current `main`, validation fails for `publish` with `E_MATRIX_INVALID` and this detail:

```text
job depends on needs-derived matrices "build" and "test", but a job can depend on only one deferred matrix expansion
```

Each deferred upload owns a separate downstream set, so the compiler rejects the join rather than upload it twice.

## What

The same workflow now passes validation. After both producers succeed, one deferred upload expands two `build` instances and two `test` instances, plus one `publish` job that waits for all four. The producers remain in the initial upload. When all four matrix instances succeed, `publish` runs once.

Intersecting downstream sets merge transitively under one deferred upload. Sharing only an upstream producer does not merge independent matrices; outside prerequisites remain in the initial upload.

Bind every producer to its exact instance key and plan digest, and count each downstream job once against the component's budget. Replays require every uploaded key to match its plan or skipped-placeholder binding. A verified failed, skipped, or cancelled producer skips only its forward closure; healthy branches still expand. Missing manifests fail without uploading jobs.

Serial deferred matrices remain unsupported, and this does not broaden environment, concurrency, or runner behavior. Hosted Buildkite join/cancellation/replay probes have not run.

Fixes [PB-3337](https://linear.app/buildkite/issue/PB-3337).


zhming0:
Released in [v0.77.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.77.0)

PR #503, opened 2026-09-16
PB-3335: Reject changed producer results in later matrix stages

zhming0:
## Why

Serial output-derived matrices now work on `main` through #508, including `plan -> build matrix -> summarize -> deploy matrix`. This PR retains the earlier-producer verification that #508 did not carry over from the original #503.

After the first matrix expands, a later compile step currently trusts the recorded rows without re-reading the earlier producer. It can upload `deploy` even if `plan`'s manifest is missing or belongs to a new attempt with identical output.

Save this complete workflow as `.github/workflows/serial-matrices.yml`. It uses only shell commands; the build and deploy steps print their selected matrix values.

```yaml
name: Serial matrices
on: push

jobs:
  plan:
    runs-on: ubuntu-latest
    outputs:
      matrix: ${{ steps.emit.outputs.matrix }}
    steps:
      - id: emit
        run: echo 'matrix=[{"target":"amd64"},{"target":"arm64"}]' >> "$GITHUB_OUTPUT"

  build:
    needs: plan
    runs-on: ubuntu-latest
    strategy:
      matrix:
    steps:
      - run: echo "Build ${{ matrix.target }}"

  summarize:
    needs: build
    runs-on: ubuntu-latest
    outputs:
      matrix: ${{ steps.emit.outputs.matrix }}
    steps:
      - id: emit
        run: echo 'matrix=[{"region":"us-east"},{"region":"eu-west"}]' >> "$GITHUB_OUTPUT"

  deploy:
    needs: summarize
    runs-on: ubuntu-latest
    strategy:
      matrix:
        include: ${{ fromJSON(needs.summarize.outputs.matrix) }}
    steps:
      - run: echo "Deploy ${{ matrix.region }}"
```

Both current `main` and this PR validate the workflow successfully:

```sh
buildkite-gha validate .github/workflows/serial-matrices.yml
```

The previous `E_MATRIX_INVALID` diagnostic is gone. The remaining failure is at runtime: let the first generated matrix step upload both `build` instances and `summarize`, then retry `plan` before the generated `deploy` matrix step runs. Current `main` never re-reads `plan`, so it still uploads both deploy instances using the earlier recorded build rows. It also uploads them when the earlier `plan` manifest is unavailable.

## What

Record the verified producer's instance key and canonical result digest with each resolved matrix. Before any later upload, re-read earlier producers through the existing verified result transport and compare their digests, including their attempt identities. A changed earlier result exits 1 before uploading anything:

```text
buildkite-gha: upload: earlier matrix "build" producer result changed after expansion
Retry the whole build to expand this matrix again. If the matrix producer job was retried, only a new build can expand it.
```

Missing or invalid earlier manifests also fail before any upload; they are not skips. With unchanged results, the workflow keeps #508's staged behavior:

1. The initial upload creates `plan` and a compile step that owns `build`, `summarize`, and `deploy`.
2. After `plan` succeeds, that step reads its verified output and uploads two `build` instances (`amd64`, `arm64`), `summarize` depending on both, and a second compile step waiting for `summarize`.
3. After `summarize` succeeds, the second compile step reads its verified output and uploads two `deploy` instances (`us-east`, `eu-west`). No earlier jobs are uploaded again.

Commit the workflow before using `upload`: generated compile steps reopen it from the build checkout. Further serial stages repeat the same process while retaining ownership of every remaining downstream job.

The rebase keeps #508's shared `upload --stage-digest` compile path and #502's join support. Earlier plans, workflow sources, action locks, token authority, budgets, and child-step replay checks remain intact.

No new graph shapes, environment, concurrency, or runner behavior are added. Workflow concurrency with output-derived matrices remains rejected. Each matrix producer must have exactly one statically known instance; existing output, matrix, and graph limits still apply. No live Buildkite serial-stage proof was run.

Fixes [PB-3335](https://linear.app/buildkite/issue/PB-3335/expand-output-derived-matrices-across-serial-workflow-stages).


mcncl (approved):

zhming0:

PR #508, opened 2026-09-17
PB-3335: Recursively compile deferred workflow stages

zhming0:
Linear: PB-3335

## Why

A matrix can already come from a job output. But when the job that produces the output is itself expanded by a deferred upload, the second matrix is rejected:

```yaml
jobs:
  plan:
    outputs: { matrix: ${{ steps.matrix.outputs.matrix }} }
    # ...
  build:
    needs: plan
    strategy:
      matrix:
    # ...
  package:
    needs: build
    outputs: { targets: ${{ steps.targets.outputs.targets }} }
    # ...
  deploy:
    needs: package
    strategy:
      matrix:
        include: ${{ fromJSON(needs.package.outputs.targets) }}
    # ...
  release:
    needs: [publish, deploy]
```

```
x [E_MATRIX_INVALID] matrix values come from a job output that exists only after that job runs; ... {job=deploy}
  detail: a needs-derived matrix cannot depend on a job that is itself expanded by a deferred upload
```

## What

The workflow compiles. The deferred upload runs as a chain of stages, and every stage is the same `upload` command:

```
Result: compilable
! [W_MATRIX_DEFERRED] matrix values come from output "matrix" of job "plan"; jobs "build", "package",
  "deploy", "publish", "release" are compiled and uploaded by step "gha-build-matrix" after that job
  finishes ...; the matrices of jobs "deploy" come from jobs that upload compiles, so later steps
  expand them in turn
```

### One `upload` command, one compile path

The `continue` subcommand is gone. The importer and every deferred step are stages of one compilation, and a deferred step runs the internal form of the same command:

```sh
"$distribution" upload --stage-digest sha256:<digest> --stage-producer <job-id>
```

Both forms build one `hostedCompileRequest` (`stageRecord.compileRequest`) and send it to the same hosted compiler that `validate` uses. After compiling, both call `stageRecord.advance`, which does everything that follows a compile: checks the recompiled graph against the recorded one (drift), checks that the one matrix still deferred is the next stage the importer promised, counts the budget, writes skipped placeholders, and writes the next stage's record and step through `stageRecord.write`. The importer is a stage whose record exists only in memory (`importerStage`); a child stage is the parent record with the resolved rows, the current graph, and the remaining budget replaced. The compiler has no notion of a stage: a matrix whose producer is deferred is recorded as a later matrix of the producer's continuation, and once that producer's rows are supplied it is an ordinary deferred root.

What is specific to a deferred step is orchestration only (`upload_stage.go`): download and verify the record, read the producers' verified results, skip the roots whose producer failed, and upload the pipeline or reproduce it on replay.

Before this change the importer and `continue` had separate post-compile code (`checkDrift`, `checkContinuationsUnchanged`, `deferredJobs`, `nextContinuation`, `buildContinuationArtifacts`) that had to agree with each other. `advance` replaces them, so the CLI stage code shrinks (1,463 → 1,389 lines) while gaining chaining.

### Persisted state

The stage record (`buildkite-gha/stage/v1`, at `.buildkite-gha/stages/<digest>.json`) keeps only what a later Buildkite job cannot recompute: the compile request, the graph, the rows earlier stages resolved, the continuation it owns, the others it must leave unchanged, and the budget. Producer bindings are no longer stored; a stage derives them from the graph and the decoder verifies each producer is in the graph with a matching logical job and plan digest. The event snapshot lives at `.buildkite-gha/stages/events/<digest>.json` and is always the importer's; stages never copy it.

### Command transition

`continue` was only ever emitted into generated pipelines, and a deferred step downloads and runs the exact executable digest its own importer uploaded, so a build never mixes a `continue`-emitting importer with a stage-form executable. The plugin and runner repositories do not reference `continue`. `docs/cli.md` records the transition. Old-form records are rejected by the version check, the same as any other compiler-version mismatch.

### Capability

- One `gha-build-matrix` step still owns the whole deferred component. When it runs, it expands `build`, compiles `package` and `publish`, and appends a second step that waits on `package` and expands `deploy` and `release`. Any number of stages chain this way.
- Every component gets its full share of the graph bound at every depth; a stage passes what it did not spend to the next one (`jobs uploaded so far + remaining budget == component share`). The row check before compilation counts later stages' placeholders and dependents, so a producer output cannot starve the stages after it.
- Skips and failures carry across stages: if `plan` fails, every job in the component gets a skipped placeholder; if only `package` fails, the earlier stage's jobs run and the later stage skips `deploy` and `release`.
- Replaying a stage reproduces the same steps, including the next stage's step and record digest, so a retried upload is idempotent.
- Ownership and joins from #502 are unchanged: a shared dependent still appears once, uploaded by the stage that resolves its last root.

Tests cover two- and three-stage chains, staged skips and failures, budgets at each stage, replay of every stage, malformed stage records, and the `--stage-digest` form's option exclusivity.

Builds on the joins from #502 (merged). Supersedes #503.


zhming0:
@buildsworth-bk this is largely a refactor PR, there is no new behavior to any existing green builds, it only bring some red builds to green. I don't think it's spicy enough to need human. Review again

PR #510, opened 2026-09-18
Avoid flaky cancellation tests while Docker calls start

zhming0:
## Why

The fake Docker subprocess creates its calls file before acquiring the write lock. Polling can read that zero-byte file and fail while decoding JSON, as happened in build #2416 during `TestRunHostJobServiceReadinessCancellationCleansEverything`.

## What

Treat an empty calls file as no calls yet. A deterministic regression test covers the window before the subprocess's first write.


zhming0:

PR #512, opened 2026-09-20
Separate action graph construction from invocation analysis

mcncl:
## Why

Action compilation mixes source resolution and graph construction with credential analysis for each invocation. This makes source lifetimes and input-dependent decisions harder to reason about independently.

## What

Build a completed, source-independent action graph before analysing invocation inputs. Graph construction owns and releases source leases; analysis uses the loaded metadata and programs. Existing caller boundaries, diagnostics, and credential decisions stay unchanged.

Add regression coverage for analysis after source removal, changing inputs, and provider guards, plus a repeatable composite-action benchmark. Analysis remains sequential. This does not add graph caching or claim a speedup.


wolfeidau (approved):
👍🏻

zhming0:

PR #513, opened 2026-09-21
Avoid rebuilding repeated actions during workflow compilation

mcncl:
## Why

Matrix jobs and repeated `uses` steps rebuild the same actions and their dependencies during validation and plan construction, even when source resolution is already cached.

## What

Reuse completed action graphs within one bundle compilation when the ordered references, workspace, and workflow source match. Each invocation still gets fresh credential analysis, and each plan owns its mutable data. Failed graphs and graphs from earlier compilations are not reused.

In the local benchmark with eight matrix jobs and three repeated composite-action steps per job, compilation drops from about 15.3 ms to 4.5 ms, with allocations falling from roughly 46,900 to 8,900. Network-heavy builds may see different gains.


wolfeidau (approved):
👍🏻

zhming0:

PR #514, opened 2026-09-21
Keep action programs unchanged during credential analysis

mcncl:
## Why

Credential analysis writes expression annotations into shared action programs through a shallow copy. With completed action graphs now reused, this makes an otherwise read-only operation depend on exclusive access to the graph.

## What

Derive annotations and inspect event usage in one traversal of a private action copy. Input precedence, credential decisions, and diagnostic ordering stay unchanged. Bundle compilation remains sequential, and plans keep their independent copies.

The regression test checks the in-memory annotations that JSON comparisons omit, including concurrent readers with different inputs and providers and a failed child lookup.


lox (approved):

zhming0:
Released in [v0.80.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.80.0)

PR #517, opened 2026-09-21
PB-3363: Accept boolean reusable-workflow inputs from job outputs

zhming0:
## Why

Hermes CI passes `deps: ${{ needs.detect.outputs.deps == 'true' }}` to a boolean reusable-workflow input. Buildkite rejects this valid workflow with `deferred reusable-workflow input "deps" must be string`.

The importer only retained string templates for inputs that depend on earlier job outputs, and the runtime rendered every result as a string. Removing the compile-time rejection alone would make false values truthy in downstream conditions.

## What

Carry the declared boolean or number type through compilation, nested calls, and immutable job plans. Evaluate these inputs as typed expressions against verified producer outputs and reject mismatched results. Existing string plans keep their rendering behavior.

Regression coverage exercises the original boolean comparison through nested workflow compilation, plan encode/decode, output resolution, and actual job execution. False and missing outputs skip the job; true runs it. Additional cases cover numeric values, type mismatches, schema validation, and expression-analysis consistency.

Fixes [PB-3363](https://linear.app/buildkite/issue/PB-3363/accept-typed-reusable-workflow-inputs-derived-from-needs-outputs) in the Buildkite GitHub Actions compatibility project. Replaces the closed, unmerged [Hermes workaround PR](https://github.com/buildkite-playground/hermes-agent/pull/2); no Hermes workflow changes are needed for this fix.

The original Hermes workflow now passes graph construction and matrix expansion unchanged. Its separate runner support/mapping failures remain outside this change.


zhming0:
Released in [v0.82.0](https://github.com/buildkite/buildkite-gha/releases/tag/v0.82.0)

